PR #5, opened 2017-02-02
Artwork fix

### LINKED ISSUE #4: atvremote artwork command should save the artwork
Currently it seems `atvremote artwork` doesn't put the artwork anywhere.

robbiet480:
The `artwork` command now saves the artwork to `artwork.png` if the artwork is available.

Closes #4

### DIFF
--- a/pyatv/__main__.py
+++ b/pyatv/__main__.py
@@ -158,6 +158,15 @@ def _handle_command(args, loop):
             _print_commands('Metadata', metadata)
             _print_commands('Playing commands', playing, newline=False)
 
+        elif cmd == 'artwork':
+            artwork = yield from _exec_command(atv.metadata, 'artwork', *cmd_args)
+            if artwork is not None:
+                file_ = open('artwork.png', 'w')
+                file_.write(artwork)
+                file_.close()
+            else:
+                print('No artwork is currently available.')
+
         elif cmd in ctrl:
             yield from _exec_command(atv.remote_control, cmd, *cmd_args)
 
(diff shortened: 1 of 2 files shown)


postlund:
I haven't even thought about this for the reference application, but it's a good idea. One idea that I've had with atvremote is to avoid hardcoing use of the public API as much as possible, so that it automatically adjusts to new commands automatically. If you for instance add a new method to ``pyatv.interface.RemoteControl``, that command will be available in atvremote with no further changes. So, to keep it that way, a suggestion from my side is to add a ``-o`` flag that redirects output to a file instead. It's not very useful for most commands, but it keeps my vision for atvremote and it also makes it possible to specify filename. Thoughts?

robbiet480:
I don't know if this will be useful at all since the only thing that outputs non-human-readable data is the artwork command no?

pyatv/__main__.py
```diff
@@ -158,6 +158,15 @@ def _handle_command(args, loop):
             _print_commands('Metadata', metadata)
             _print_commands('Playing commands', playing, newline=False)
 
+        elif cmd == 'artwork':
+            artwork = yield from atv.metadata.artwork()
+            if artwork is not None:
+                file = open('artwork.png', 'wb')
```
postlund: I would recommend to use the ``with`` statement for opening tile file as that closes it no matter what happens.

postlund:
Yeah, you are right. Currently it might a bit overkill so I can accept this solution. But I would prefer if you fix my comment before merging.

Issue #16, opened 2017-02-04
Play media by specifying a URL


postlund:
From what I have gathered, it should be possible to use the ``play`` command and pass a URL to the device to play it. Some inspiration can be found in these projects:

https://github.com/hellozimi/AirPlayer/blob/master/airplayer.py
https://github.com/Tomohiro/airplayer

Also the Unoffical AirPlay Protocol Specification:

https://nto.github.io/AirPlay.html#video-httprequests

I would like to add the same functionality to pyatv. Suggested API is ``pyatv.interface.RemoteControl.play(url)``.

postlund:
I have this mostly working, but I am stopped by aiohttp at the moment. To receive events, I need a reverse tunnel and to set that up I need to include an "Upgrade" header in the request. This does not seem to work with aiohttp. Created an issue for it here: https://github.com/KeepSafe/aiohttp/issues/1587

postlund:
All done!

PR #30, opened 2017-02-07
Play media by specifying a URL

postlund:
This shall be the final version for now. I will have to make this better later and add more tests, but at least I have a base now.


postlund:
This fixes #16.

PR #37, opened 2017-02-09
Rename topmenu to top_menu

### LINKED ISSUE #34: Remote control commands design
Is it voluntary to use `topmenu` rather than `top_menu` to name  the remote control command? It does not match the design of `set_position`.

This is not really important but as it is a fresh project, there is still time to change :)

pierrickrouxel:
Fixes #34


postlund:
Thanks! 🐬

PR #56, opened 2017-03-08
More robust AirPlay error handling

postlund:


postlund:
I don't like this work-around and I'm gonna close it. If issue persists I will have to come up with a better solution.

PR #62, opened 2017-03-14
Support for shuffle and repeat modes

postlund:


postlund:
Verified on my device.

PR #105, opened 2017-08-15
Fix AirPlay session close and update AirPlay play_state logic

alexeykhit:


pyatv/airplay/player.py
```diff
@@ -39,42 +42,43 @@ def play_url(self, url, position=0):
...
                 address, headers=headers, data=body, timeout=TIMEOUT)
-            yield from self._wait_for_media_to_end()
+            yield from self._wait_for_media_to_end(session)
         finally:
             if resp is not None:
                 resp.close()
+            # Apple TV 3 Gen have a bug. It stays in playing state forever
+            # after stream finished. We need close session every time when
+            # video stops to fix it.
+            session.close()
```
postlund: It's strange that it would behave like that (I haven't seen it on my ATV3). AFAIK the underlying connection should be closed when close() is called. Do you have any reference media that I can try?

alexeykhit: Like this. I use it with HomeAssistant.

```python
import requests
requests.post(
    'http://localhost:8123/api/services/playlist_player/play_media',
    json={
        'entity_id': 'media_player.apple_tv',
        'media_content_type': 'video',
        'media_content_id': 'http://clips.vorwaerts-gmbh.de/big_buck_bunny.mp4'
    })
```

I used same logic with my pure python realization:

`requests stream=True` to keep connection open. And `r1.content` at the end to close it.

```
status = 'loading'

r1 = requests.post(
    'http://%s:7000/play' % host,
    data='Content-Location: %s\nStart-Position: 0.0\n\n' % url,
    stream=True
)

while status != 'stopped':
    time.sleep(1)

    r2 = requests.get('http://%s:7000/scrub' % host)

    if status == 'loading':
        if 'position: 0' not in r2.text:
            status = 'playing'
    elif status == 'playing':
        if 'position: 0' in r2.text:
            status = 'stopped'

r1.content
```

But then I switched to HomeAssistant with your library.

postlund:
Sorry for dragging this out, I'm quite busy at the moment. I will look into it on sunday and next week as I get a lot of free time then.

Is it the API call to Home Assistant that blocks now btw?

alexeykhit:
I did not understand the question.

postlund:
I was referring to your comment:
```
# Apple TV 3 Gen have a bug. It stays in playing state forever
# after stream finished. We need close session every time when
# video stops to fix it.
```
So, I assume that something you do is getting blocked because the connection is not closed? The reason I'm questioning the change at all is because it's recommended to re-use the same ``ClientSession`` throughout the application life time. This is mainly because it has expensive resources (e.g. a thread pool) associated with it. It might also contain special configuration setup by the user of the library, like proxy servers and so on, which would not be respected when creating a new ``ClientSession`` instance every time.

alexeykhit:
Yes. I made a playlist_player component for Home Assistant. It can play YouTube playlists (and other sources) on Apple TV (and other devices).

Component monitors the state of devices. And with this bug it does not know when to start next video on Apple TV.

postlund:
I've been thinking about it a bit more and even though this is not an ideal solution, I think we will have to go with it for now. So if you fix so that it works again (I have moved files around a bit when refactoring but I haven't changed the behavior of airplay) I'll merge it for you. Also, I would prefer if you add/modify an existing test to cover the uncovered statement.

postlund:
I will close this and use #118 instead.

PR #109, opened 2017-09-03
WIP: Initial changes to more flexible config

postlund:
This PR will make the config used when connecting to a device or scanning for devices more flexible. **It is a breaking change!**

Still not finished, a few more cases to handle and additional tests. New API is not final yet, it will probably change a bit before merge.


postlund:
OK to lower code coverage now before MRP is in place. Also, documentation is out-of-date for now. I will fix that later.

PR #137, opened 2018-02-21
Deprecate python 3.4

postlund:
From now on python 3.5.3 is needed. Also python 3.7 should work fine. Documentation is not updated yet, I will make a big overhaul later.


postlund:
Had to leave out py3.7 for now.

Issue #205, opened 2019-10-07
Update documentation for 0.4.0


postlund:
**Short feature/function description**
Most of the documentation is currently wrong and not up to date with current implementation. Since I use GitHub for development, I would also like to migrate and use as many services from GitHub as possible to simplify the process. So the next step is to migrate to GitHub pages.

**What needs to be done?**
Migrate current documentation in `docs/` to GitHub pages and update what is currently not correct. Also add some information about MRP, but refer to https://github.com/jeanregisser/mediaremotetv-protocol as much as possible.

**Is this a breaking change?**
Code-wise no, but the old readthedocs page will be removed

**Anything else worth knowing?**
No

postlund:
Old documentation should be moved from `docs` to for instance `docs/old` so that GitHub can use `docs` for GitHub pages. I'm also considering using the wiki. Not sure which is best really.

postlund:
I have disabled read the docs and moved the old documentation. So it's time to start migrating the old docs to GitHub pages and use markdown instead.

postlund:
After lots of trial-and-error, there's a place holder page at https://postlund.github.io/pyatv/

PR #258, opened 2019-11-14
Add more documentation

postlund:
Documentation is still far from complete, but I have migrated some parts and written some new stuff.


postlund:
Part of #205.

PR #262, opened 2019-11-17
Add MRP functionality

ckeehan:
Add MRP functionality and update documentation with changes

### DIFF
--- a/pyatv/mrp/__init__.py
+++ b/pyatv/mrp/__init__.py
@@ -1,7 +1,7 @@
 """Implementation of the MediaRemoteTV Protocol used by ATV4 and later."""
 
 import logging
-import time
+import asyncio
 from datetime import datetime
 
 from pyatv import (const, exceptions)
@@ -53,7 +53,7 @@ class MrpRemoteControl(RemoteControl):
         if lookup:
             await self.protocol.send(
                 messages.send_hid_event(lookup[0], lookup[1], True))
-            time.sleep(lookup[2])
+            await asyncio.sleep(lookup[2])
             await self.protocol.send(
                 messages.send_hid_event(lookup[0], lookup[1], False))
         else:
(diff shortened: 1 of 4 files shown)


pyatv/mrp/__init__.py
```diff
@@ -48,6 +53,7 @@ def __init__(self, loop, protocol):
         if lookup:
             await self.protocol.send(
                 messages.send_hid_event(lookup[0], lookup[1], True))
+            time.sleep(lookup[2])
```
postlund: Two things here:
* Why is this needed?
* You must use the asyncio version (asyncio.sleep), this will block the event loop

ckeehan: Good catch. I just committed a change to implement asynchronous sleep.

The messages.send_hid_event(arg, arg, True) sends the keydown event to Apple TV, messages.send_hid_event(arg, arg, False) sends keyup. For normal use you want these events to execute back to back with no delay, but a sleep allows for sending "press and hold" button events. Specifically implemented here with the home_hold command.

postlund: Ah yes, that is correct. I admit that I did not read my own code good enough here... But it looks good to me!

postlund:
Travis has some complaints, just fix those and it should be ok to merge. I guess you just have to copy from interface to dmap more or less.

ckeehan:
Sorry for the delay. I believe those commits take care of the changes necessary.

postlund:
No worries, do you know if the commands in the DMAP-implementation actually works? Otherwise I was just thinking they could throw NotSupportedError for now.

ckeehan:
Excellent point. Unfortunately I have no way of testing DMAP so I've updated it to throw a NotSupportedError

postlund:
Looks good! 👍

Issue #246, opened 2019-11-08
Add smoke test for atvremote


postlund:
**Short feature/function description**
I unintentionally break `atvremote` every now and then because I don't have any regression tests for it. I want to add that.

**What needs to be done?**
I intend to reuse mocks from integration tests (aiozeroconf and scan mock). This way I can configure a static simulated environment that atvremote will "see". I will also use one of the fake devices so I can try connecting to it and getting basic information.

Some things to test:

* Scanning
* Pairing
* Push updates (if not too cumbersome)
* Playing information
* Manual connect

**Is this a breaking change?**
No

**Anything else worth knowing?**
I intend to make the tests simple to write and more or less only verify output to stdout:

```python
def test_scan(self):
    atvremote("scan")
    self.assertIn("aa:bb:cc:dd:ee:ff", self.stdout)
```

postlund:
I've dismissed this issue a bit but I'm starting to feel that it is important and that I should add it to the 0.4.0 milestone. So I'm just gonna do that.

postlund:
Definition of Done for this story:
* Test for MRP pairing
* Test for MRP verifying metadata (`atvremote ... plyaing`)
* Test for play something with AirPlay
* Test manual connect

postlund:
Pairing MRP is more or less the same as AirPlay from atvremote'a point-of-view. So I'm gonna skip that for know as it saves me some time now, not having to implement pairing in the fake device.

PR #263, opened 2019-11-17
atvremote CLI: fixed credentials reference to .service.credentials

michaelrcarroll:
Tiny tiny fix!

Was receiving this error using the example atvremote CLI to pair:
```
root@nuc:/opt/pyatv# atvremote --id 00:11:22:33:44:55 --protocol mrp pair
Enter PIN on screen: 0123
Pairing seems to have succeeded, yey!
ERROR: Pairing failed
  File "/usr/local/lib/python3.7/dist-packages/pyatv/__main__.py", line 127, in pair
    await self._perform_pairing(pairing)
  File "/usr/local/lib/python3.7/dist-packages/pyatv/__main__.py", line 163, in _perform_pairing
    pairing.credentials))
AttributeError: 'MrpPairingHandler' object has no attribute 'credentials'
```

It needed an update to reference the correct ```PairingHandler.service.credentials``` attribute.

### DIFF
--- a/pyatv/__main__.py
+++ b/pyatv/__main__.py
@@ -160,7 +160,7 @@ class GlobalCommands:
         if pairing.has_paired:
             print('Pairing seems to have succeeded, yey!')
             print('You may now use these credentials: {0}'.format(
-                pairing.credentials))
+                pairing.service.credentials))
         else:
             print('Pairing failed!')
 


postlund:
Good catch, this again reminds me of why I need #246 😉 Also, I need to add `service` to the `PairingHandler` interface because this relies on private implementation right now. I want it to work like this so it's not a problem

michaelrcarroll:
Agreed! #246 would go a long way towards flagging things like this in the future. I'm just getting started integrating this into my home automation, but if I continue to use it, I'll see if I can spare some cycles to contribute.

Issue #243, opened 2019-11-07
Add persistent storage for device data


postlund:
**Short feature/function description**
Currently `pyatv` doesn't store any configuration on disk. It would be useful to have simple persistent storage of credentials after pairing, to not have to deal with long command lines. At first this storage can hold credentials and identifiers, but in the future other stuff could be stored as well. So it should be genetic enough to support that.

**What needs to be done?**
* Add a simple storage module
* Integrate in `atvremote`
* Update documentation 

**Is this a breaking change?**
No

**Anything else worth knowing?**
No

postlund:
Fixed by #2151!

PR #300, opened 2019-12-02
Assumes protocol if none given

vegerot:
Seems like a simple addition.  Right now, if the user tries to pair without a protocol, it will immediately exit.  Now, it will at least try to guess one, and if not it'll exit—giving the same result as before.

The way this ends up working out is that it prefers remotes to airplay.  Since protocol 1 is mutex to 2, don't need to worry about preferring one of those.  And if it has 2 and 3, it'll default to the remote.  If it has just 3, it'll use just 3, etc.

### DIFF
--- a/docs/documentation/atvremote.md
+++ b/docs/documentation/atvremote.md
@@ -69,6 +69,7 @@ Which protocols a device supports can be seen with `scan`. But in general you ne
 either `mrp` (devices running tvOS) or `dmap` (Apple TV 3 and earlier). If you also want to
 stream video, you can pair `airplay` as well. The procedure is the same for all of them, just
 change the argument provided to `--protocol`.
+If no protocol is specified, it will assume `mrp` (devices running tvOS) or `dmap` (Apple TV 3 and earlier) first, or AirPlay if that's the only one available.
 
 ### Credentials
 
--- a/pyatv/__main__.py
+++ b/pyatv/__main__.py
@@ -121,12 +121,9 @@ class GlobalCommands:
 
         if self.args.protocol is None:
             logging.warning('No protocol specified')
-            try:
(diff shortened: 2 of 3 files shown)


postlund:
I think we can make this behavior default in `atvremote`, but it also needs to be documented. Added some comments you can have a look at, let me know if you agree. Documentation-wise, the "Pairing with a device" chapter should be updated:

https://github.com/postlund/pyatv/blob/master/docs/documentation/atvremote.md

One thing to not about this change (as it is harder to reason about then it looks like). When pairing and no protocol is given, `atvremote` will default to to one of `DMAP` or `MRP` depending on which is available and which was found first and `AirPlay` last because of how of my AirPlay-hack. This is fine today since devices that support both protocols will become more and more rare due to tvOS 13. A nice addition here would be to pick the next protocol without credentials, since now you would always pair the same protocol over and over again. 

The result obtained from `apple_tv.services` (or `apple_tv.get_services()` in your case) does not have any fixed order since it's a dict. From python 3.7 it is fixed to insertion-order (earlier versions is unspecified) but since services can be discovered in any order, you can make no assumptions on which protocol you find first. So you could end up pairing AirPlay too.

Combining this feature while respecting credentials (i.e. not pairing protocols which already has credentials) and #243 would be a nice change in the future. New credentials can be saved to file and loaded automatically next run, so you always pair the next protocol without credentials.

Sorry about the long post here, this is an interesting issue and I kinda dodged these parts because I knew they were tricky 😉

vegerot:
@postlund Did you leave comments in a review, or are you just talking about this post?  If the former, how do I access those comments?

postlund:
I managed to not post the review comments... But they are up now!

postlund:
Getting old, closing this.

PR #321, opened 2019-12-10
Add initial smoke tests for atvremote

postlund:
Only scanning and AirPlay pairing is currently tested. This will be
the start. Once more MRP tests are finished so the fake MRP device
(which this tests rely on) are usable, I will add metadata tests.


postlund:
Implements parts of #246.

PR #349, opened 2020-01-04
Initial support for unicast scanning

postlund:
This is a proof-of-concept for supporting unicast scanning, i.e. scanning for specific hosts by providing their IP-addresses. To test this with `atvremote`, pass the `--scan-hosts` flag:

```shell
$ atvremote --scan-hosts 192.168.1.10 scan
```

It is possible to pass multiple addresses as well, comma-separated:

```shell
$ atvremote --scan-hosts 192.168.1.10,192.168.1.16 scan
```
 
It can be used with regular commands and provides some speed-up since no waiting has to be done for multicast responses:

```shell
$ time atvremote --scan-hosts 10.0.10.81 -n Vardagsrum playing 
Media type: Unknown
Device state: Idle

real 0m0.620s
user 0m0.520s
sys  0m0.056s
```


postlund:
Latest commit should be pretty "ready", but documentation is missing. Will write that soon.

PR #394, opened 2020-01-21
Update troubleshooting.md

alistairg:
Update to docs to include PIN challenges :)

### DIFF
--- a/docs/support/troubleshooting.md
+++ b/docs/support/troubleshooting.md
@@ -14,10 +14,26 @@ change something.
 
 Please see [this](../scanning_issues/) page on scanning issues.
 
+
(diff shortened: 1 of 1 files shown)


docs/support/troubleshooting.md
```diff
@@ -14,11 +14,27 @@ change something.
 
 Please see [this](../scanning_issues/) page on scanning issues.
 
+
+### I can't enter leading zeros for my PIN
```
postlund: This is instructions for the Home Assistant component and will have to be added to the Home Assistant documentation later. Here we should only care for `pyatv` stuff.

postlund:
Opened #588 with updates about no PIN code being displayed (since I cannot push to this PR).

Issue #268, opened 2019-11-20
Provide means of determining supported features of a device


postlund:
**Short feature/function description**
It's currently not possible to programmatically determine which features a device supports. You have to call a function and see if it raises an exception or not. That is not ideal and must be fixed as it's hard to implement good applications otherwise.

I need to think more about how the interface should look as I'm not sure yet. So some investigation is included here as well.

**What needs to be done?**
TBD

**Is this a breaking change?**
No (this is a new interface)

**Anything else worth knowing?**
Since the feature set is more or less equal for `DMAP` and `MRP`, today this is vital now. But as the feature gap expands it is a must have. So it needs to implement it quite early.

Issue #249, opened 2019-11-12
Add type hints to public API


postlund:
**Short feature/function description**
Type checking is already done with mypy (limited use right now). It would be good to add type hints at least to the public parts of the API, extending it further as time goes by. Shouldn't take much effort.

**What needs to be done?**
The public API is basically defined in these files:

* `__init__.py`
* `conf.py`
* `convert.py`
* `exceptions.py`
* `helpers.py`
* `interface.py`

Look them over and add type hints as needed.

**Is this a breaking change?**
No
**Anything else worth knowing?**
No

PR #458, opened 2020-02-23
MRP Power State support

stickpin:


pyatv/mrp/player_state.py
```diff
@@ -105,6 +106,12 @@ def _add_listeners(self):
         self.protocol.add_listener(
             self._handle_set_now_playing_client,
             protobuf.SET_NOW_PLAYING_CLIENT_MESSAGE)
+        self.protocol.add_listener(
```
postlund: This will not work as it is only sent during connection set up and handled by that code (see comment in `protocol.py`).

pyatv/mrp/protocol.py
```diff
@@ -166,8 +166,8 @@ def message_received(self, message, _):
                 self._outstanding[identifier].semaphore, message)
             self._outstanding[identifier] = outstanding
             self._outstanding[identifier].semaphore.release()
-        else:
-            self._dispatch(message)
+
+        self._dispatch(message)
```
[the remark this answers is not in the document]
stickpin: @postlund, you are right, because of the initial DeviceInfoMessage I've changed the `self._dispatch(message)`.

Can you give an example please how can I store the message in protocol.py?

pyatv/mrp/protobuf/__init__.py
```diff
@@ -124,6 +124,7 @@
 CLIENT_UPDATES_CONFIG_MESSAGE = ProtocolMessage.CLIENT_UPDATES_CONFIG_MESSAGE
 CRYPTO_PAIRING_MESSAGE = ProtocolMessage.CRYPTO_PAIRING_MESSAGE
 DEVICE_INFO_MESSAGE = ProtocolMessage.DEVICE_INFO_MESSAGE
+DEVICE_INFO_UPDATE_MESSAGE = ProtocolMessage.DEVICE_INFO_UPDATE_MESSAGE
```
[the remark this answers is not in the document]
stickpin: Not able to generate. Getting the following error, no matter what I do:
```
  File "scripts/protobuf.py", line 259, in <module>
    sys.exit(main())
  File "scripts/protobuf.py", line 249, in main
    return update_auto_generated_code()
  File "scripts/protobuf.py", line 154, in update_auto_generated_code
    subprocess.run(["protoc",
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/subprocess.py", line 489, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/subprocess.py", line 1702, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'protoc'
```

pyatv/mrp/player_state.py
```diff
@@ -165,3 +172,14 @@ def playing(self):
 
             if self.listener:
                 await self.listener.state_updated()
+
+    async def _handle_device_info_message(self, message, _):
```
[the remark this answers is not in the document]
stickpin: Once I am removing async, I am hitting this error:
```
2020-02-24 14:11:24 ERROR: Failed to handle message
  File "/Volumes/Archive/Jacob/Development/Github/pyatv/pyatv/mrp/connection.py", line 106, in data_received
    self._handle_message(data)
  File "/Volumes/Archive/Jacob/Development/Github/pyatv/pyatv/mrp/connection.py", line 120, in _handle_message
    self.listener.message_received(parsed, data)
  File "/Volumes/Archive/Jacob/Development/Github/pyatv/pyatv/mrp/protocol.py", line 170, in message_received
    self._dispatch(message)
  File "/Volumes/Archive/Jacob/Development/Github/pyatv/pyatv/mrp/protocol.py", line 178, in _dispatch
    asyncio.ensure_future(
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/tasks.py", line 673, in ensure_future
    raise TypeError('An asyncio.Future, a coroutine or an awaitable is '
TypeError: An asyncio.Future, a coroutine or an awaitable is required
```

With async everything fine.

pyatv/__main__.py
```diff
@@ -254,6 +254,11 @@ def playstatus_error(_, exception):
         """Inform about an error and restart push updates."""
         print("An error occurred (restarting): {0}".format(exception))
 
+    @staticmethod
+    def powerstate_update(_, powerstate):
+        """Print what is current power state when it changes."""
+        print(str(powerstate), flush=True)
```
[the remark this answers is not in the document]
stickpin: I actually followed your example: 
```
    def playstatus_update(_, playstatus):
        """Print what is currently playing when it changes."""
```
Let me know what you want me to put there. :)

pyatv/mrp/protobuf/__init__.py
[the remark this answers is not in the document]
postlund: You need protoc in your PATH as that is used by the script. If you wait just a bit until I manage to fix and merge #460, the script can download it for you:

```shell
$ python3 scripts/protobuf.py --download generate
```

This way you also get the "correct" version.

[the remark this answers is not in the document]
postlund: This PR also fixes the mentioned bug, so `DEVICE_INFO_UPDATE_MESSAGE` will be present after it has been merged.

pyatv/mrp/protocol.py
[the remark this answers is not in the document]
postlund: The response (`DeviceInfoMessage`) is returned by `send_and_receive` here:

https://github.com/postlund/pyatv/blob/070946fd05d9c84bbfa2f3f0c98af62a459ec9e3/pyatv/mrp/protocol.py#L68

So you can just save it like (also add `self.device_info = None` in constructor to make pylint happy):

```python
    self.device_info = await self.send_and_receive(msg)
```

It is then available via `protocol.device_info` from where ever you have a reference to the active `protocol` instance. It is still the matter of adding an API for accessing this though. Maybe we can do it like this...

Add a new interface in `interface.py` called `Power` (as I mentioned before) looking something like this:

```python
class PowerListener:
    """Listener interface for power updates."""

    @abstractmethod
    def powerstate_update(self, old_state, new_state):
        """Device power state was updated."""
        raise NotImplementedError()


class Power:
    __metaclass__ = ABCMeta

    def __init__(self):
        """Initialize a new AppleTV."""
        self.__listener = None

    @property
    def listener(self):
        """Object receiving power state updates.

        Must be an object conforming to PowerListener.
        """
        return self.__listener

    @abstractproperty
    def power_state(self):
        """Return device power state."""

    @abstractmethod
    def turn_on(self):
        """Turn device on."""

    @abstractmethod
    def turn_off(self):
        """Turn device off."""

# At the bottom in `AppleTV`, add the following:

    @abstractproperty
    def power(self):
        """Return API for power management."""
```

Since we don't yet know how to manage power state with pre-tvOS devices, we need a way to work around that. So my suggestion here is that we also introduce a new state type, which is also more flexible in case we want to add new states in the future. One suggestion is to add the following to `const.py`:

```python
class PowerState(Enum):
    """All supported power states."""

    Unknown = 0
    Off = 1
    On = 2
```

We can hardcode to always return `Unknown` for `DMAP` and let it be either `On` or `Off` for `MRP` depending on value of `logicalDeviceCount`. The `turn_on` and `turn_off` methods can be implemented in any way that fits (basically what you do in the component now). If we find better ways of doing the same thing later, we can improve that.

I also added a listener for power state as well as interface for it (the interface will mostly be relevant when I implement #249). This way we can keep all power related logic in the same place.

Next part is to implement `MrpPower` in `pyatv/mrp/__init__.py` that implements the interface above and satifies the abstract members. This is very similar to how all the interfaces are implemented, `MrpMetadata` can be good to look at as inspiration. I imagine that you can move the message listening logic (`protocol.add_listener` to this class as well). You also need to create an instance of `MrpPower`. This is done in `__init__` in the same file, again you can look at how it's done for `MrpMetadata`.

The same thing also needs to be implemented for `DMAP` (i.e. add `DmapPower` to `pyatv/dmap/__init__.py`, etc). We can work on that over time. We also need to consider tests and documentation in due time.

There's a lot to it here, so I understand if it's a lot to take in. I will help you as much as I can 😄 You basically add completely new API with this, so it pulls a lot of infrastructual code with it. But it's a very good exercise. To not be overwhelmed, do one thing at the time, e.g. focus on `MRP` for now, fix `DMAP`, next, tests after that and finish off with documentation. That makes it a lot easier.

We need to add a few things in `atvrmote` as well, so we can test the feature. Most things are handled automatically, but new interfaces must be added manually at the moment (there's room for improvement here). It follows the same pattern more or less. Do the same thing as for `Metadata` but replace with `Power` in these places:

https://github.com/postlund/pyatv/blob/070946fd05d9c84bbfa2f3f0c98af62a459ec9e3/pyatv/__main__.py#L77
https://github.com/postlund/pyatv/blob/070946fd05d9c84bbfa2f3f0c98af62a459ec9e3/pyatv/__main__.py#L92
https://github.com/postlund/pyatv/blob/070946fd05d9c84bbfa2f3f0c98af62a459ec9e3/pyatv/__main__.py#L493
https://github.com/postlund/pyatv/blob/070946fd05d9c84bbfa2f3f0c98af62a459ec9e3/pyatv/__main__.py#L507

Then you can use `atvremote` and access the API like so:
```shell
$ atvremote ... power_state
$ atvremote ... turn_on
$ atvremote ... turn_off
```

pyatv/__main__.py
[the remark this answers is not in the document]
postlund: I was thinking about the output text 😄 

```python
    print("Power state changed to", powerstate, flush=True)
```

postlund:
There's a big lack in developer documentation for `pyatv` at the moment. I intend to use the wiki here for instructions, but I haven't come around for it yet. I'll see if I can get going on the today or tomorrow.

pyatv/mrp/protobuf/__init__.py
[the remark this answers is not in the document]
stickpin: Merged PR [#460](https://github.com/postlund/pyatv/pull/460) into my fork and now everything generated automagically. :)

Thanks!

stickpin:
@postlund Just to keep you posted.
I think I've already implemented most of the things. I missing some points but I will ask you later on.

I will do commit later today, so you will be able to review it tomorrow morning. :)

postlund:
@stickpin Sounds good, I will be ready 😄 Time to go to bed now though...

stickpin:
@postlund please check the new [commit](https://github.com/postlund/pyatv/pull/458/commits/e46656647db9f26e18298282d942da13bf4491d6) and let me know what you think.

All the changes are currently relevant to MRP, I will work on DMAP later on.
`atvremote` commands are already working as well, except listener.

and here where I need your help... I am a bit lost with this listener (PowerListener).
Can you please explain a little bit more about how to use this `PowerListener`? 

Thanks in advance! :)

pyatv/interface.py
```diff
@@ -410,6 +410,43 @@ def connection_closed(self):
...
+    @abstractmethod
+    def powerstate_update(self, old_state, new_state):
+        """Device power state was updated."""
+        raise NotImplementedError()
+
+class Power:
+    __metaclass__ = ABCMeta
+
+    def __init__(self):
+        """Initialize a new AppleTV."""
```
postlund:
```suggestion
        """Initialize a new Power instance."""
```

pyatv/mrp/__init__.py
```diff
@@ -326,6 +326,68 @@ def artwork_id(self):
...
+        super().__init__()
+        self.loop = loop
+        self.protocol = protocol
+        self.remote = remote
+        self._power_state = None
+        self._listener = None
+        self._add_listeners()
+
+    def _add_listeners(self):
+        self.protocol.add_listener(
```
postlund: You can remove this as it will never happen in this context.

pyatv/interface.py
```diff
@@ -410,6 +410,43 @@ def connection_closed(self):
...
+        """
+        return self.__listener
+
+    @abstractproperty
+    def power_state(self):
+        """Return device power state."""
+        raise exceptions.NotSupportedError()
+
+    @abstractmethod
+    def turn_on(self):
```
postlund: Ah, this was bad on my end. This method should be async:
```suggestion
    async def turn_on(self):
```

pyatv/interface.py
```diff
@@ -410,6 +410,43 @@ def connection_closed(self):
...
+        """Turn device on."""
+        raise exceptions.NotSupportedError()
+
+    @abstractmethod
+    def turn_off(self):
```
postlund: Same here, `async` as well.

pyatv/mrp/__init__.py
```diff
@@ -326,6 +326,68 @@ def artwork_id(self):
...
+            current_power_state = await self.update_power_state(self.protocol.device_info, protobuf.DEVICE_INFO_MESSAGE)
+            return current_power_state
+        else:
+            return self._power_state
+
+    def power_state(self):
+        """Return device power state."""
+        return self._get_current_power_state()
+
+    def turn_on(self):
```
postlund:
```suggestion
    async def turn_on(self):
```

pyatv/mrp/__init__.py
```diff
@@ -326,6 +326,68 @@ def artwork_id(self):
...
+        """Turn device on."""
+        return self._turn_on()
+
+    def turn_off(self):
```
postlund:
```suggestion
    async def turn_off(self):
```

pyatv/mrp/__init__.py
```diff
@@ -326,6 +326,68 @@ def artwork_id(self):
...
+        else:
+            return self._power_state
+
+    def power_state(self):
+        """Return device power state."""
+        return self._get_current_power_state()
+
+    def turn_on(self):
+        """Turn device on."""
+        return self._turn_on()
```
postlund: You can inline this method now:

```python
async def turn_on(self):
    """Turn device on."""
    await self.remote.menu()
```

The pattern I've used where I return a coroutine in non-async methods are bad practice, so I will try change them to async as we go along.

pyatv/mrp/__init__.py
```diff
@@ -326,6 +326,68 @@ def artwork_id(self):
...
+
+    def turn_off(self):
+        """Turn device off."""
+        return self._turn_off()
```
postlund: Inline here as well.

pyatv/mrp/__init__.py
```diff
@@ -326,6 +326,68 @@ def artwork_id(self):
...
+    
+        if logicalDeviceCount >= 1:
+            self._power_state = PowerState.On
+        elif logicalDeviceCount == 0:
+            self._power_state = PowerState.Off
+        else:
+            self._power_state = PowerState.Unknown
+        
+        _LOGGER.debug('Power state is now %s', self._power_state)
+        return self._power_state
```
postlund: From an abstraction perspective, it's better if you split this method in two, since it's a callback method that is not supposed to return anything. Would sugget this instead;
```python
    async def update_power_state(self, message, _):
        self._power_state = self._get_power_state(message)
        _LOGGER.debug('Power state is now %s', self._power_state)

    @staticmethod
    def _get_power_state(device_info):
        logicalDeviceCount = device_info.inner().logicalDeviceCount
        if logicalDeviceCount >= 1:
            return PowerState.On
        elif logicalDeviceCount == 0:
            return PowerState.Off
        return PowerState.Unknown
```
You can then re-use the latter method above.

pyatv/mrp/__init__.py
```diff
@@ -326,6 +326,68 @@ def artwork_id(self):
...
+        await self.remote.home_hold()
+        await self.remote.select()
+
+    async def _get_current_power_state(self):
+        if self._power_state is None:
+            current_power_state = await self.update_power_state(self.protocol.device_info, protobuf.DEVICE_INFO_MESSAGE)
+            return current_power_state
+        else:
+            return self._power_state
+
```
postlund: You need to add `@property` here.

pyatv/mrp/__init__.py
```diff
@@ -326,6 +326,68 @@ def artwork_id(self):
...
+            protobuf.DEVICE_INFO_UPDATE_MESSAGE)
+
+    async def _turn_on(self):
+        await self.remote.menu()
+
+    async def _turn_off(self):
+        await self.remote.home_hold()
+        await self.remote.select()
+
+    async def _get_current_power_state(self):
```
postlund: You can inline this method.

pyatv/mrp/__init__.py
```diff
@@ -326,6 +326,68 @@ def artwork_id(self):
...
+        if self._power_state is None:
+            current_power_state = await self.update_power_state(self.protocol.device_info, protobuf.DEVICE_INFO_MESSAGE)
+            return current_power_state
+        else:
```
postlund: No need for `else` here (you will get a lint warning about it).

pyatv/__main__.py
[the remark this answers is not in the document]
stickpin: Not relevant anymore.

pyatv/mrp/player_state.py
stickpin: Not relevant anymore.

postlund:
Some further improvements we can do later is to check if power state changes when calling `turn_on`/`turn_off` according to expected state and raise an exception if that isn't the case. So, if the user calls `turn_off` and we don't get a `DeviceInfoMessage` with `logicalDeviceCount` set to 0 (I.e. power state is off) within a second or so, then the action probably failed. It's good if we can signal this back to the user. But that can be a separate PR.

stickpin:
@postlund, just did a [commit](https://github.com/postlund/pyatv/pull/458/commits/0b9bae2c12459191c32e32d6c77348b044acb1c9) with the listener and few more fixes.
Can you please take a look again and give your feedback? :)

postlund:
I'll have a look soon and get back to you before bedtime :+1:

pyatv/interface.py
```diff
@@ -434,7 +434,7 @@ def listener(self):
         return self.__listener
 
     @abstractproperty
-    def power_state(self):
+    async def power_state(self):
```
postlund: This should be left as non-async. Properties  should generally not perform any intensive computations, just return a stored value.

pyatv/mrp/__init__.py
```diff
@@ -339,6 +339,13 @@ def artwork_id(self):
         return MrpPlaying(self.psm.playing)
 
 
+class PowerListener(PowerListener):
```
postlund: No need for this class here as it's not used (only users outside of `pyatv` provides an implementation of this). The only reason for the `PowerListener` interface is so that it's easy to see which methods that should be implemented.

pyatv/mrp/__init__.py
postlund: Still valid comment.

pyatv/mrp/__init__.py
```diff
@@ -359,16 +367,21 @@ def __init__(self, loop, protocol, remote):
...
-        return self._power_state
+        return self._current_power_state
 
     @property
-    def power_state(self):
+    def listener(self):
+        """Object receiving power state updates."""
+        return self.__listener
+
+    async def power_state(self):
```
postlund:
```suggestion
    def power_state(self):
```

pyatv/mrp/__init__.py
```diff
@@ -359,16 +367,21 @@ def __init__(self, loop, protocol, remote):
...
+        if self._current_power_state is None:
+            updated_power_state = await self.update_power_state(self.protocol.device_info, protobuf.DEVICE_INFO_MESSAGE)
+            return updated_power_state
         
-        return self._power_state
+        return self._current_power_state
 
     @property
-    def power_state(self):
+    def listener(self):
```
postlund: This is OK but you also need to provide a setter property, otherwise it's not possible to assign a listener. I assumed you took some inspiration from `PushUpdater`, so you can do it in the same way:

https://github.com/postlund/pyatv/blob/024284bca4565b3d75509fa82a509dcbeeeb7a96/pyatv/interface.py#L362

pyatv/mrp/__init__.py
postlund:
```suggestion
    def _get_current_power_state(self):
```

postlund: This method is a bit ambivalent. It is named `get_` but also modifies internal state, which is a bit contradictory and I would prefer if you avoid doing that. The power state can be determined by the latest received `DeviceInfoMessage`, which was either the message received when connecting *or* the latest message received in `update_power_state`. Here we can use some python magic...

The `or` operator works with objects. For instance `"hello" or 123` returns `"hello"` and `None or "foobar"` returns `"foobar"`. We can use this logic here to pick which message to use. Lets assume `update_power_state` saves the message internally in `MrpPower` like this:

```python
def __init__(self):
    self.device_info = None  # We must set this in the constructor as well

async def update_power_state(self, message, _):
    self.device_info = message
```

To get the "latest" message, we can now do:
```python
latest_device_info = self.device_info or self.protocol.device_info
```
If we have not received a `DeviceInfoUpdateMessage`, then `self.device_info` will be `None` and we will get the value of `self.protocol.device_info` instead (initial `DeviceInfoMessage`). If we have received the message, then we will get that message instead. So the code would look something like this:

```python
def _get_current_power_state(self):
    latest_device_info = self.device_info or self.protocol.device_info
    return self. _get_power_state(latest_device_info)
```
This has the benefit of being short and does not modify any state and does not require any asyncio operation. You can even inline it into the `power_state` method.

By doing this you don't have to save the state (`_current_power_state`), just keep the `DeviceInfoMessage` instead.

pyatv/mrp/__init__.py
```diff
@@ -380,8 +393,16 @@ def power_state(self):
         await self.remote.select()
 
     async def update_power_state(self, message, _):
```
postlund: You can simplify this a bit now since `power_state` will return current power state:
```python
async def update_power_state(self, message, _):
    old_state = self.power_state
    self.device_info = message

    if self.power_state != old_state:
        _LOGGER.debug('Power state changed from %s to %s', old_state, self.power_state)
        if self.listener:
            self.loop.call_soon(self.listener.powerstate_update, self._current_power_state, self._new_power_state)
```

pyatv/interface.py
stickpin: The only problem here. If I am changing it to non-async, I am getting the following error:

```
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 572, in _run_application
    return await cli_handler(loop)
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 390, in cli_handler
    return await _handle_commands(args, config, loop)
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 482, in _handle_commands
    ret = await _handle_device_command(args, cmd, atv, loop)
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 515, in _handle_device_command
    return (await _exec_command(
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 538, in _exec_command
    value = await tmp(*args)
TypeError: object PowerState can't be used in 'await' expression

>>> An error occurred, full stack trace above
```

It seems that cli command cannot be non-async, which makes sense.

pyatv/mrp/__init__.py
stickpin: The only problem here. If I am changing it to non-async, I am getting the following error:

```
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 572, in _run_application
    return await cli_handler(loop)
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 390, in cli_handler
    return await _handle_commands(args, config, loop)
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 482, in _handle_commands
    ret = await _handle_device_command(args, cmd, atv, loop)
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 515, in _handle_device_command
    return (await _exec_command(
  File "/Volumes/Archive/Development/Github/pyatv/pyatv/__main__.py", line 538, in _exec_command
    value = await tmp(*args)
TypeError: object PowerState can't be used in 'await' expression

>>> An error occurred, full stack trace above
```

It seems that cli command cannot be non-async, which makes sense.

pyatv/mrp/__init__.py
stickpin: I hope I've finally understood the Listeners idea.
It should be fixed now.

pyatv/mrp/__init__.py
stickpin: the `new_state` was missing.

My code looks like this:
```
    async def update_power_state(self, message, _):
        old_state = await self.power_state()
        new_state = self._get_power_state(message)
        self.device_info = message

        if new_state != old_state:
            _LOGGER.debug('Power state changed from %s to %s', old_state, new_state)
            if self.listener:
                self.loop.call_soon(self.listener.powerstate_update, old_state, new_state)
```

stickpin:
@postlund please let me know if you find more things to fix. :) 
If not what are the next steps?

stickpin:
@postlund btw... which device are you using for DMAP testing?

postlund:
@stickpin Apple TV 3, but I haven't done much development for that in a long time so the fake device used by tests are usually good enough.

stickpin:
For those who following. 
I've updated an unofficial HASS plugin to support all latest changed: https://github.com/stickpin/hass-atv-beta

stickpin:
@postlund hi. :)

I've implemented the Power into the DMAP but I am not sure if it's the right way to implement it.
I also have nothing to test it with.

I've also fixed many `tox` complaints. :) except for this one:
`pyatv/dmap/__init__.py:445:0: W0223: Method 'power' is abstract in class 'AppleTV' but is not overridden (abstract-method)` which I guess related to the wrong implementation from my side.

One more thing, once I am running `tox` i am getting the following error:
`ERROR: InvocationError for command /Volumes/Archive/Development/Github/pyatv/.tox/lint/bin/pylint pyatv examples (exited with code 4)`

Thanks in advance!

stickpin:
Heh... I was a bit too fast with the conclusions about this `suspend` and `wakeup` commands.
It seems that `suspend` command not sending CEC to the TV so it stays on. :(
Rolling back the change.

postlund:
@stickpin Looks good so far, no comments from me 😄 The last pylint error is because you missed adding the `power` property in `DmapAppleTV` (you have the the corresponding method for `MRP` here as reference: https://github.com/postlund/pyatv/pull/458/files#diff-47265af3491d37ee8fe3cc93b6f0682aR509). The `InvocationError` will disappear once you fix the pylint error, so nothing to worry about there.

Yeah, `wakeup` and `suspend` doesn't do what you expect 😛 Should really deprecate those...

Regarding testing `DMAP`. The easiest way is probably to write a functional test for it. Since it doesn't really do anything, you can just check that the state is set to what is expected and call `turn_off` and `turn_on` while expecting exceptions. I can give you a kick-start for `power_state`. Open `tests/dmap/test_dmap_functional.py` and add (and import `PowerState` from `const` at the top):

```python
    @unittest_run_loop
    async def test_unsupported_power_state(self):
        self.assertEqual(await self.atv.power.power_state, PowerState.Unknown)

        # Call turn_on/off and check for exception
```
You can look at `test_not_supportedt` (feel free to remove `t` at the end) in the same file for inspiration.

postlund:
Just wanted to mention that the error you see in the log is a known issue and nothing you need to worry about. Good job on getting the code in place! :+1:

https://github.com/postlund/pyatv/wiki/Submitting-a-PR#known-issues

stickpin:
@postlund, there's an option for me to re-run these tests?
In general, I am planning to work on tests today and docs.
Should we already tag it as a candidate for 0.4.1 release? :)

postlund:
@stickpin Yeah, if you go to the failing job you should have a dropdown menu with "Re-run jobs" in the top-right corner. You can do it from there (assuming there's no strange permissions things).

Sure, we can tag it for 0.4.1 😄 I won't finish the remaining stuff before you so it might as well be there.

stickpin:
hehe... here where it's becoming tricky. :))
I am absolutely lost how to even start implementing tests for MRP.

How should I set an initial `PowerState`, for example, to `PowerState.Off` and how subscribe to the `listener` so I will able to get the status change when I am pressing `turn_on` for example?

One more question, in `cli` once I run `bin/atvremote commands` I see the following for Power commands:
```
Power commands:
 - listener - Object receiving power state updates
 - power_state - Return device power state
 - turn_off - Turn device off
 - turn_on - Turn device on
```

How do I hide listener? :))

I will work on docs meantime.

stickpin:
I think I've completed more or less with the docs.
Let me know what you think.

postlund:
@stickpin Yeah, it's a steep curve to get going there 😉 I would suggest that you start with a simple failing test and play around with the logs and try to correlate to code. For example, add this to `test_mrp_functional.py`:

```python

    @unittest_run_loop
    async def test_power_state(self):
        self.assertEqual(self.atv.power.power_state, PowerState.Unknown)
```
This should fail because the fake device sets `logicalDeviceCount` to 1:

https://github.com/postlund/pyatv/blob/1f97630c5f95c0634bb907d485a48d4c0be439f2/tests/mrp/fake_mrp_atv.py#L258

https://github.com/postlund/pyatv/blob/1f97630c5f95c0634bb907d485a48d4c0be439f2/pyatv/mrp/messages.py#L41

If you run:
```shell 
$ pytest --disable-warnings --log-level=debug -k test_power_state
```
You should see the test run with the same debug logs as when you run `atvremote` with `--debug` as well as some logs from the fake device. Take some time and get used to the output - it helps a lot. You can add your own logs with for instance `_LOGGER.debug(...)` or `logging.debug(...)`. I recommend these over `print` because the latter is printed separately by `pytest`, so it's hard to know how it correlates to the other print outs.

So, I suggest that you make two tests: one by just checking `power_state` and one with a listener. Let's assume default state of the device is "on" when starting the test (which it is now). Then `power_state` should return that it is on, so you check for that. Then you call `turn_off` and again check `power_state`, but this time it should be off. Then you finish off with `turn_on` and verify that `power_state` is on.

There's a small quirk here that you need to know. Because everything here happens over a TCP connection (like when talking to a real device), it might take some time before all messages has been sent and received properly. To workaround this, you have to poll `power_state` several times until it reaches the desired state (with a timeout in case it never happens). I have a helper method for this when polling `playing`, e.g. waiting for the device to play something with a particular title or a button to be pressed:

https://github.com/postlund/pyatv/blob/1f97630c5f95c0634bb907d485a48d4c0be439f2/tests/common_functional_tests.py#L126

You can do the same thing, but check `self.atv.power.power_state == PowerState.XXX`. This will keep checking if the condition is fulfilled with a sleep between each check. After a couple of seconds it will fail with an exception in case it fails.

Another tricky part is to figure out in the fake device if it is in sleep mode or not (I.e. if `logicalDeviceCount` should be 0 or 1). The fake device currently savesthe latest button press here:

https://github.com/postlund/pyatv/blob/1f97630c5f95c0634bb907d485a48d4c0be439f2/tests/mrp/fake_mrp_atv.py#L286

A simple solution that works in this case would be to set a flag internally if "select" is pressed and the previously pressed button was "home_hold". That way we can hide logic from the test (I.e. do not check for this key press combination in the test), which makes the test easier. You should also send a DeviceInfoUpdateMessage here when implementing listener support (do that when writing the device test).

I hope this can give you a start, I'll continue to check in when I can but weekend is family time! Will have a look at the documentation ASAP.

And oh, we need to filter "listener" in `atvremote`. I have missed that. You can try to fix in if you want, otherwise I'll do it.

stickpin:
@postlund I've added MRP tests for initial/turn_on/turn_off power states.
Now only listener left. Hopefully will have enough time to understand and close it tomorrow.

Btw, I don't have an option to re-run the tests, I guess it's related to some permissions issue.

postlund:
@stickpin Looks good so far 😊 Will be able to do a full review tomorrow. Regarding the listener, have a look at the the tests for push updates. They should be quite similar. More or less implant a listener that saves the input arguments and the you verify them in the test.

Ok, good to know. I'll re-trigger when necessary.

postlund:
Comment about `turn_on`

stickpin:
@postlund let me know what you think about this solution.
This one should cover the callback case. :)

postlund:
@stickpin This looks good 😊 I suggest that you also add checks for `new_state`. It is also bast practice in the world of python to never define new variables outside of the constructor of a class (it causes less confusion). So just add a constructor and set `old_state` and `new_state` to `None`. I haven't checked the log but I think pylint or flake8 nags about that as well.

I will do a big review of everything tonight and see if I can play around with it a bit. Want to see if `WakeDevice` is viable here.

stickpin:
> @stickpin This looks good 😊 I suggest that you also add checks for `new_state`. It is also bast practice in the world of python to never define new variables outside of the constructor of a class (it causes less confusion). So just add a constructor and set `old_state` and `new_state` to `None`. I haven't checked the log but I think pylint or flake8 nags about that as well.

It's actually funny because I've had a constructor but then I've deleted it because everything worked without it and pylink was not complaining about it. So I've just done Command+Z to return it back. :))
I also added `new_state` validation.

> I will do a big review of everything tonight and see if I can play around with it a bit. Want to see if `WakeDevice` is viable here.

`wakeup` command was working fine when I was testing it and HDMI CEC was triggered, so yes, in theory, we can switch from `menu` to `wakeup`, but the problem with `suspend` stays.

pyatv/mrp/__init__.py
```diff
@@ -339,6 +339,64 @@ def artwork_id(self):
...
+
+    @property
+    def power_state(self):
+        """Return device power state."""
+        currect_power_state = self._get_current_power_state()
+        return currect_power_state
+
+    async def turn_on(self):
+        """Turn device on."""
+        await self.remote.menu()
```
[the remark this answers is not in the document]
stickpin: @postlund  Yep, it worked! :)
Will take a look, maybe we have something similar for Standby? :)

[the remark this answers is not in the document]
postlund: Good 😄 Nope, nothing similar for standby I'm afraid. It's done indirectly via some other non-obvious message in that case.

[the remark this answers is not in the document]
stickpin: Checking MRP private headers [here](http://developer.limneos.net/index.php?ios=13.1.3&framework=MediaRemote.framework) maybe will lead to some direction.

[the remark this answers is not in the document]
postlund: I'm *pretty* sure there nothing similar.

docs/development/listeners.md
```diff
@@ -66,3 +66,24 @@ is connection oriented. It's another case for `DMAP`, since that protocol is
...
+e.g. the device turned on or turned off. The API is defined by the
+`interface.PowerListener` interface and works similarly to how push updates works.
+
+Here is a simple example:
+
+```python
+class PowerListener:
+
+    def powerstate_update(self, old_state, new_state):
+        print('Power state changed from %s to %s', old_state, new_state)
```
postlund: C-style is only supported by `logging`, here you should use `.format`. When I decide to not support python 3.5 anymore, we can use f-strings instead.

docs/development/power_management.md
```diff
@@ -0,0 +1,38 @@
...
+
+```python
+await pwrc.turn_on()
+await pwrc.turn_off()
+```
+
+To get current power state use following function:
+
+```python
+def power_state(self):
```
postlund: You should add `@property` above this line since this is a property.

pyatv/mrp/messages.py
```diff
@@ -42,6 +42,36 @@ def device_information(name, identifier):
     return message
 
 
+def device_information_update(name, identifier, logical_device_count):
```
postlund: Maybe we can do some code sharing with `device_information` here? I generally not a fan of passing bool-flags, but maybe we can live with that here (i.e. pick correct message type from a flag)?

tests/mrp/fake_mrp_atv.py
```diff
@@ -283,6 +283,20 @@ def handle_send_hid_event(self, message, inner):
...
         elif down_press == 0:
             if (use_page, usage) in self.outstanding_keypresses:
+                if (
+                    _convert_key_press(use_page, usage) == 'select'
+                    and self.last_button_pressed == 'home'
+                   ):
+                    msg = messages.device_information_update(
+                        'pyatv', message.identifier, 0)
+                    self.send(msg)
+                if (
```
postlund: I think this part is no longer valid, right? There should also be some logic handling `WakeDevice`?

stickpin:
@postlund hopefully fixed all the issues.
Let me know what you think. :)

postlund:
Digged a bit more into the codecov issues and it has to do with secrets in forked repos when using GitHub actions. Forked repos cannot see secrets in the main repo, so it fails because of that. We will just have to hope that we haven't introduced any regressions on other platforms. Don't think it will since you don't use any platform specific stuff.

stickpin:
@postlund great! :)
The next step is to update the HASS Plugin. :)
I think it may be tricky because the plugin should support DMAP and MRP at the same time and behavior of turn_on/turn_off actions should be different. I will see how to implement it.

postlund:
@stickpin Yeah, I really need to make a push for getting the PR merged into Home Assistant so that we can work towards that version. What's in the custom component isn't up-to-date, so making changes there will just make it hard to merge back later.

My intention is to start working on #268 soon (like today or so), which would provide some mechanism to determine if various features are available or not. This would effectively solve the main issue, since there would be an API function that we can check. The best solution I think is still to add an option that is opt-in. By "option" I mean adding it here:

https://github.com/postlund/hass-atv-beta/blob/fc79765a8d854903df0c8417f3fc164d72c3fb11/custom_components/apple_tv/config_flow.py#L404

stickpin:
@postlund can I ask you to release lastest commits together with MRP Power State as some kind of v0.4.1a1 or something... So I will not need to install custom dependencies and I will be able to use one from the pip... :)

postlund:
@stickpin I can do a pre-release later this evening.

postlund:
@stickpin Pre-release 0.4.1a1 is available now.

stickpin:
@postlund thanks a lot!
I've implemented an option in the config flow to enable the new power management functionality.
https://github.com/stickpin/hass-atv-beta

Once HASS will push an update, I will submit a PR with this functionality.

PR #471, opened 2020-02-26
gha: Test on windows-latest

postlund:


postlund:
Not sure why this doesn't work as intended, can't find that anyone else has this problem with GitHub Actions.

PR #476, opened 2020-03-02
Add device information interface

### LINKED ISSUE #220: Expose more device information
Currently `pyatv` doesn't have any information about the device it is communicating with (OS version for instance). This information is not communicated via the protocols. However, AirPlay expose OS version, device model and MAC-address amongst other thing. So this information can be stored and exposed via an interface.

**What needs to be done?**
* Store mentioned data from zeroconf in the AirPlay service when it is found
* Add an additional interface in `interface.py` and implement that
* Add tests

New interface but does not break APi

No

postlund:
This PR introduces an interface to retrieve various kinds device information, e.g. running OS, hardware model, etc. I think I'm ready, but I'm gonna have another look to see if I missed anything. If anyone wants to try it out, please give it a spin!

Fixes #220.

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -58,6 +58,8 @@ page_links:
       title: Start
     - link: /development/scan_pair_and_connect/
       title: Scan, Pair and Connect
+    - link: /development/device_info/
+      title: Device Information
     - link: /development/control/
       title: Control
     - link: /development/power_management/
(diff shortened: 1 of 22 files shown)


postlund:
One thing I came to think of: there's no way to get a summaryof the device information in `atvremote` (only specific things, like operating system: `atvremote ... operating_system`). It would be convenient to add a command that prints everythin, e.g. `atvremote ... device_info`

PR #484, opened 2020-03-03
if: Add support for play_pause

### LINKED ISSUE #442: Add play_pause function to remote control API
It's convenient to have a toggle API for play pause. MRP has native support for it, not sure about DMAP. Would have to investigate that. Fallback should be to look at device state and call play/pause depending on state.

**What needs to be done?**

* Add new function to `interface.RemoteControl`
* Implement for MRP and DMAP
* Write tests
* Update documentation

No

No

postlund:
Fixes #442.


postlund:
Just verified in the Music app and it works as expected!

PR #488, opened 2020-03-05
if: Add interface for supported features

### LINKED ISSUE #268: Provide means of determining supported features of a device

postlund:
Here is the intended interface for working with supported features. By brainstorming I came up with four states of a feature. It is...

* not supported by the device at all (`FeatureState.Unsupported`)
* supported but there is no way to figure out if it is possible to "use" now, e.g. most features pre-tvOS (`FeatureState.Unknown`)
* supported but currently not possible to use, e.g. it is not possible to pause if nothing is playing (`FeatureState.Unavailable`)
* supported and possible to use, e.g. play/pause when watching something

In some cases a feature can support additional options (e.g. skip time), so I have included options for that as well. Might not implement it for now, but this way I can keep the API intact when I decide to do so.

This API is intended to be used like this:

```python
features = atv.features

if features.get_feature(FeatureName.Play).state == FeatureState.Available:
    await atv.remote_control.play()
else:
    print("Not possible to play now")
```

All features are available via the method `Features.all_features` as well (with an argument that allows to include unsupported features as well). I might add some convenience methods once I start working with the API.

NB: This is WIP, so there are still things missing. Here is a rough checklist:

- [ ] Volume support in `MRP` (currently don't know status of this)
- [x] Investigate if I can do anything better in `DMAP`
- [x] Tests (atvremote and functional)
- [x] Documentation
- [x] `atvremote` support
- [ ] General clean up after I'm done with everything (e.g. make sure I've added typing)

Fixes #268.


postlund:
My initial concept doesn't really work, so I will change it a bit.

PR #508, opened 2020-03-18
protobuf: Workaround for older versions

postlund:
The only reason a newer version of protobuf is needed is because of
print_unknown_fields when logging. Check if this flag is supported and
use it if it is, otherwise ignore it so older versions of protobuf can
be used.


postlund:
Main reason why this is needed is because Home Assistant pins protobuf 3.6.1 due to various components needing that. So it clashes with `pyatv`, requiring 3.8.0 or later.

PR #547, opened 2020-03-31
flush output of each line printed by atvscript

dfrommi:
Output-flushing is required to avoid buffering between processes.

Without flush, client applications only receive the output from atvscript
when either a bigger amount of output is generated or the application
terminates.

### DIFF
--- a/pyatv/scripts/atvscript.py
+++ b/pyatv/scripts/atvscript.py
@@ -33,11 +33,14 @@ class PushPrinter(PushListener):
 
     def playstatus_update(self, updater, playstatus: Playing) -> None:
         """Inform about changes to what is currently playing."""
-        print(self.formatter(output_playing(playstatus, self.atv.metadata.app)))
+        print(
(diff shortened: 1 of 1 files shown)


postlund:
Yes, you are absolutely right. There are two print's in appstart that you can flush as well (and fix linting by running `black .`). Otherwise good to go!

dfrommi:
Thanks, done.

postlund:
Looks good 👍 Looks like there's some problem with Actions, lets see if it get fixed during the day. Otherwise you can try to force-push again and see if that helps.

postlund:
I have no idea why there's so much problems with GitHub Actions that I haven't seen before. Created a new PR, #552, with the same content to see if it works better. It's so strange.

postlund:
#552 merged correctly, so the changes are in. Sorry about the mess and long lead time here!

dfrommi:
Thanks a lot

PR #560, opened 2020-04-06
mrp: Add error codes

postlund:
Here are finally all error codes in `ProtocolMessage`. They are very verbose and not named for easy development. I will clean naming up later, just wanted to check them in to simplify debugging in case an error occurs.


crxporter:
Can the error codes be sent with status:error to atvscript?

PR #618, opened 2020-04-24
airplay: Use default timeout for HTTP requests

postlund:


postlund:
The reason for this commit is to allow `play_url` to wake up a device.

PR #677, opened 2020-06-01
Added Telia Play to playbackRate issue

KibosJ:
Added Telia Play to the list of app known to incorrectly report "playbackRate"


postlund:
👍

PR #696, opened 2020-06-05
Clean up home key in MRP implementation.

ckeehan:
The HID key the Remote app sends for `home` and `home_hold` is different than what is used in the MRP implementation here. Functionality remains the same just updating to send the same key the Remote app does.

### DIFF
--- a/pyatv/mrp/__init__.py
+++ b/pyatv/mrp/__init__.py
@@ -59,7 +59,7 @@ _KEY_LOOKUP = {
     "select": [1, 0x89, 0],
     "menu": [1, 0x86, 0],
     "topmenu": [1, 0x86, 1],
-    "home": [12, 0x40, 1],
+    "home": [12, 0x60, 1],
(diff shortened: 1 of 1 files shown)


postlund:
Looks good! You need to make the corresponding update to the MRP service in the fake device since the tests fails. I would also like it if you make it a bit more clear in the commit message what you changed, I usually look at them when writing the change log so it helps a lot. If possible, I try to prefix with a submodule name or something to easier group changes, e.g. "mrp: Correct home button HID code" or something.

postlund:
Just wanted to see if you want to finish off this PR or if I should take over and get it done?

postlund:
Now I'm starting to wonder that is correct once again regarding this... At the moment `top_menu` sends the same HID-code as you change `home` to. I will have to re-evaluate this again to make it right.

postlund:
I don't think this is relevant anymore.

PR #704, opened 2020-06-08
mrp: Update protobuf for language options.

lucasec:
From experimentation, the format of the `availableLanguageOptions` and `currentLanguageOptions` fields in the `ContentItem` protobuf are slightly different.

This introduces a new protobuf `LanguageOptionList` to try and capture the correct format for the `availableLanguageOptions` field.

This should pave the way to better understanding these fields. I'll open up an issue later with some documentation on the `EnableLanguageOption`/`DisableLanguageOption` command. The payload for that command is actually the marshal object for the `LanguageOption` run through `NSKeyedArchiver`. Go figure 🤷

### DIFF
--- a/pyatv/mrp/protobuf/ContentItem.proto
+++ b/pyatv/mrp/protobuf/ContentItem.proto
@@ -3,12 +3,18 @@ syntax = "proto2";
 import "pyatv/mrp/protobuf/ContentItemMetadata.proto";
 import "pyatv/mrp/protobuf/LanguageOption.proto";
 
+message LanguageOptionGroup {
+  optional bool allowEmptySelection = 1;
+  optional LanguageOption defaultLanguageOption = 2;
+  repeated LanguageOption languageOptions = 3;
+}
+
 message ContentItem {
   optional string identifier = 1;
   optional ContentItemMetadata metadata = 2;
   optional bytes artworkData = 3;
   optional string info = 4;
-  repeated LanguageOption availableLanguageOptions = 5;
+  repeated LanguageOptionGroup availableLanguageOptions = 5;
   repeated LanguageOption currentLanguageOptions = 6;
 //  optional Lyrics lyrics = 7;
 //  repeated Sections sections = 8;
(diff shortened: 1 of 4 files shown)


postlund:
Ok, so, now I'm looking at this. The message definitions should look something along the lines of:

```
message LanguageOptionGroup {
  optional bool allowEmptySelection = 1;
  optional LanguageOption defaultLanguageOption = 2;
  repeated LanguageOption languageOptions = 3;
}

message ContentItem {
  ..
  repeated LanguageOptionGroup availableLanguageOptions = 5;
  ..
}
```

This should stick to the same naming convention as Apple. We can keep `LanguageOptionGroup` together with ContentItem for now as it's not used anywhere else. There are a couple of places where messages should be moved back to it's parent message that I haven't fixed yet. Unless it's shared, lets keep messages together. The `Common.proto` file is something I created to share common types with.

lucasec:
Interesting, will update with the new defs. I thought the language option types might be shared with CommandInfo, but those don't actually accept the raw protobufs—they expect it wrapped in a bplist instead. So it should be safe to keep this out of the common file for now.

lucasec:
Updated for new protobufs. It seems like we may be able to pull `LanguageOption` in to `ContentItem` as well. Let me know if that's desired.

postlund:
Great, now it looks good 😊 I suspect that you are right regarding `ContentItem` based solely on naming, but I will double check that later today and get back to you. If that's the case, then we can move stuff to `Common.proto`. `CommandInfo` should be correct as-is I think. Might be missing some new field though.

lucasec:
Cool. I'm also about to open up an issue with a few more details on the langue option commands. I suspect `CommandInfo` is still mostly right.

lucasec:
Filed https://github.com/postlund/pyatv/issues/720 to hopefully fill in the rest of the details :)

postlund:
Ok, so, `availableLanguageOptions` in `ContentItem` should be `LanguageOptionGroup` (repeated as-is). `currentLanguageOptions` looks correct as it is right now.

postlund:
Should we perhaps try to make mentioned changes and try to merge this PR?

lucasec:
Sorry, lost track of this. I think the PR is actually correct as-is (the changes you mentioned should have been made in my most recent push). I re-base on latest master and re-push now though.

postlund:
Ah, you are right! I missed that. Then I can gladly merge this 👍

Issue #750, opened 2020-07-06
Use of aiozeroconf causes issues in Home Assistant


jjlawren:
**Describe the bug**
Having `aiozeroconf` installed in the same environment as Home Assistant leads to various `aiohttp` errors on unrelated components: https://github.com/home-assistant/core/issues/37303

**To Reproduce**
Have the latest `apple_tv` custom component installed on Home Assistant 0.112.0+.
**System Setup (please complete the following information):**
 - OS: Ubuntu 19.10
 - Python: 3.8.3
 - pyatv: 0.6.0
 - apple_tv custom_component: https://github.com/postlund/hass-atv-beta/tree/d2a081ead8deb82896bb634d03ac5b4a1676262a
- Apple TV: AppleTV 4

bdraco:
```

from homeassistant.components import zeroconf
zeroconf_instance = await zeroconf.async_get_instance(hass)

```

If you want to avoid `aiozeroconf` and can use the sync one, you can get the above to get the main instance that is running in HA.

You'll need the following in your `manifest.json`

```
  "after_dependencies": [
    "zeroconf"
  ],
```

postlund:
I actually did use the synchronous version before, but switched to `aiozeroconf` due to bad results. Now I have implemented parts of zeroconf myself to circumvent limitations in the libraries that prevents me from implementing certain features. I will not be able to finish that part any time soon (I.e. not dropping the dependency) and I'm quite hesitant to switch yet another time. So, I'm trying to understand the root cause. It must be something in `aiozeroconf`, right? Or does it have to do with two libraries trying to act as mdns responders at the same time?

bdraco:
It is something in `aiozeroconf` for sure.    Both of them should be able to run together just fine (with the performance penalty of running two zeroconf implementations at once -- this can be significant).  It looks like `aiozeroconf` is stomping on open file descriptors somehow.

bdraco:
Can you provide some more detail on what is broken in `zeroconf`?

postlund:
[...]

PR #756, opened 2020-07-13
scan: Migrate back to zeroconf from aiozeroconf

postlund:
As aiozeroconf causes problems (due to bugs) and is not maintained
anymore, it's better to move back to zeroconf again. Since service
discovery is already done by pyatv, only service publishing is really
used.

Relates to #750.


postlund:
Tested this by pairing with an Apple TV 3, which worked (apart from a bug I found that I will fix in another commit).

PR #771, opened 2020-07-16
gha: Run tests with python 3.9

### LINKED ISSUE #762: Run Actions with python 3.9
**What feature would you like?**
Python 3.9 is not out yet but it's possible to run tests with a beta version using actions. My intention is of course to support python 3.9.

**Describe the solution you'd like**
Add 3.9 to list of python versions to test on.

**Any other information to share?**
No

postlund:
Fixes #762.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,21 +4,27 @@ on: [pull_request]
 
 jobs:
   build:
+    name: >-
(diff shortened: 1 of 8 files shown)


postlund:
Ok, so in python 3.9 `Optional` are kept as-is and not changed to `Union[x, None]`. This is nice but kinda screws with my scripts...

PR #774, opened 2020-07-18
Resolve performance slowdown with copying playing data

### LINKED ISSUE #773: Most of the home assistant event loop is spent calling the copy in `pyatv.mrp.playing`
Screen Shot 2020-07-17 at 2 52 29 PM
Screen Shot 2020-07-17 at 2 52 19 PM

bdraco:
I'm not sure if this will cause a regression but it solves the performance issue.

Resolves #773 

Before this change the event loop was taking 13% of the recording time.  
https://maui.koston.org:8123/local/ec.svg

After its below the measurable threshold.
https://maui.koston.org:8123/local/ec3.svg

### DIFF
--- a/pyatv/mrp/__init__.py
+++ b/pyatv/mrp/__init__.py
@@ -4,7 +4,7 @@ import math
 import logging
 import asyncio
 import datetime
-from copy import deepcopy
+from copy import copy
 from typing import Dict, List, Optional, Tuple
 
 from pyatv import conf, exceptions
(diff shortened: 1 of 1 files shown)


postlund:
A `copy` won't work as `ContentItemUpdate` and `UpdateClient` modify some of the members. This will violate the immutability promise. It would have been very good with tests for this... Will try to add that later tonight. The easiest solution is to practice copy-on-write for these fields before merging changes.

postlund:
I have tried to write a failing test for the cases I mentioned but I cannot seem to make it fail. Can't really say that I'm 100% sure why but perhaps protobuf implements decent `__copy__` routines or something? Anyways, I do feel that we can merge this then. Any problems can be dealt with later.

PR #799, opened 2020-08-15
tvOS 14 Compatibility

evanreichard:
**NOTE:** I have _not_ tested this with anything other than tvOS 14. It's likely incompatible, but I'm willing to work with whomever to test that and tweak as needed.

### New:
- Switched to SEND_BUTTON_EVENT_MESSAGE vs HID (Cleaner, IMO)
- Compiled protobuf (Used `script/protobuf.py`)

### Fixed:
- Fixed ProtocolMessage protobuf typo to allow SEND_BUTTON_EVENT_MESSAGE
- Timeouts due to lack of tvOS 14 HID ACKS
- Playback monitoring issues. Appears that tvOS 14 reports using `SET_STATE_MESSAGE` vs `SET_NOW_PLAYING_CLIENT_MESSAGE`

### Removed:
- `playback_state` playbackRate issue compensation. I believe it was the Apps fault. I can no longer reproduce the previously reported bug for Amazon Prime. It was also causing issues with my playback monitoring implementation. (Need to verify this works OK for tvOS XX < tvOS 14.


evanreichard:
Im going to revisit this. After hard restarting the Apple TV features started working on the old version.

postlund:
The switch to SEND_BUTTON_EVENT_MESSAGE would be great to see in a new PR, assuming it works with tvOS 13 as well. I've looked at it before but never took the time to verify if it works and integrate it.

PR #883, opened 2020-12-03
update MANIFEST.in, add missing file

onkelbeh:
otherwise the SDIST from Pypi is not compilable, because base_versions.txt (which is read by setup.py) would be missing.


postlund:
👍

Issue #836, opened 2020-09-17
Add support for FreeBSD


atxbyea:
**Describe the bug**
Not possible to turn on apple tv from Home assistant, logs just say : [pyatv.mrp.connection] Keep-alive not supported: FreeBSD does not support keep-alive

**To Reproduce**
Home-assistant 114.4 in a FreeBSD 12.1 jail 

**Expected behavior**
AppleTV to turn on

**System Setup (please complete the following information):**
 - OS: FreeBSD 12.1
 - Python: 3.8.5
 - pyatv: 0.7.2
- Apple TV: AppleTV 4 HD

**Additional context**

postlund:
Ok, I should add support for FreeBSD. The errors are unrelated however as that is just a warning (and means that unexpected loss of network connections won't be detected, e.g. if the Apple TV crashes or cable is pulled). Please enable additional logging and attach them here.

atxbyea:
Sorry for the late reply, this should contain the requested info.
[pyatv.log](https://github.com/postlund/pyatv/files/5282194/pyatv.log)

postlund:
At a quick glance I don't see anything strange here. When you say "not possible to turn on", what do you actually mean by that? What isn't working?

If I prepare a commit with keepalive support for FrreBSD, can you help out testing that with `atvremote`?

atxbyea:
Aye, it didn't turn on, I can play more and log more as needed, and test as needed

postlund:
Do you mean in Home Assistant (so it doesn't show any state) or do you mean turn on as in it shows something on your screen?

atxbyea:
Toggling the `remote.` entity does nothing :-)

postlund:
The integration does not support power management, so that is expected. It only toggles if the entity is on or off.

atxbyea:
Sorry, my bad, I meant the `media_player.` entity, this button 
![Screenshot_20201008-211333_Home Assistant](https://user-images.githubusercontent.com/46269073/95503198-4e246e00-09ab-11eb-97e4-1e1d5ff897f5.jpg)

postlund:
It's the same, `turn_on` and `turn_off` use the exact same logic for both `media_player` and `remote`.

atxbyea:
Good to know, so the turn on\off isn't supposed to work at all, would it be possible to remove the button somehow or is that hard coded in the device features ?
If you need me to test anything for keepalive I can do that when needed.

postlund:
[...]

PR #897, opened 2020-12-16
Use TCP keepalive on more platforms

paxswill:
This should allow using keepalive on platforms that have the appropriate constants available within Python, as well as Darwin. The main personal driver for this was to silence the warning (and actually implementing keepalive support) on FreeBSD, which should address #836. I also added a couple of tests for the `tcp_keepailve()` function.

### DIFF
--- a/pyatv/support/net.py
+++ b/pyatv/support/net.py
@@ -123,7 +123,7 @@ def tcp_keepalive(sock) -> None:
         try:
             option = getattr(socket, option_name)
         except AttributeError:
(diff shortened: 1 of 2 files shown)


postlund:
Thank you very much, this is a great addition 😄 Big plus for added tests. Would have preferred to have the socket stuff using asyncio, but since this such a small change I can live without that.

Will be included in the next release I make, hopefully quite soon. I will probably align it with Home Assistant 2021.01.0.

Issue #878, opened 2020-11-23
No devices found


wayller:
- pyatv: 0.7.4
- Apple TV: Apple TV 4K tvOS 14.2

atvremote scan --debug
```
/config # atvremote scan --debug
2020-11-23 19:40:31 DEBUG: Binding on *:5353
2020-11-23 19:40:31 DEBUG: Binding on 127.0.0.1:5353
2020-11-23 19:40:31 DEBUG: Binding on 10.0.150.25:5353
2020-11-23 19:40:34 DEBUG: Auto-discovered TV-rom at 10.0.0.102:7000 (Protocol.AirPlay)
Scan Results
========================================
```

atvremote scan --debug --mdns-debug
```
/config # atvremote scan --debug --mdns-debug
2020-11-23 19:40:40 DEBUG: Binding on *:5353
2020-11-23 19:40:40 DEBUG: Binding on 127.0.0.1:5353
2020-11-23 19:40:40 DEBUG: Binding on 10.0.150.25:5353
2020-11-23 19:40:40 DEBUG: Sending multicast DNS request to 224.0.0.251:5353 (Data=35ff012000040000000000000b5f6170706c6574762d7632045f746370056c6f63616c00000c80010b5f746f7563682d61626c65045f746370056c6f63616c00000c80010e5f6d6564696172656d6f74657476045f746370056c6f63616c00000c8001085f616972706c6179045f746370056c6f63616c00000c8001)
2020-11-23 19:40:40 DEBUG: Received DNS response from ('10.0.150.25', 5353) (Data=35ff012000040000000000000b5f6170706c6574762d7632045f746370056c6f63616c00000c80010b5f746f7563682d61626c65045f746370056c6f63616c00000c80010e5f6d6564696172656d6f74657476045f746370056c6f63616c00000c8001085f616972706c6179045f746370056c6f63616c00000c8001)
2020-11-23 19:40:40 DEBUG: Received DNS response from ('10.0.150.25', 5353) (Data=35ff012000040000000000000b5f6170706c6574762d7632045f746370056c6f63616c00000c80010b5f746f7563682d61626c65045f746370056c6f63616c00000c80010e5f6d6564696172656d6f74657476045f746370056c6f63616c00000c8001085f616972706c6179045f746370056c6f63616c00000c8001)
[...]

Issue #765, opened 2020-07-15
Not able to connect to Apple TV in different subnet


ITTV-tools:
**Describe the bug**
I can not add ATV to my Home Assistant. 
My network is separated in multiple VLANs. The ATV is in VLAN 1 and Home Assistant in VLAN 2.
Both can talk to each other. 
I use a MDNS repeater to use for example Homekit in other VLANs. 

In older versions of the integration, HA could see the Apple TV and I got the text field to enter the PIN code.
But the PIN was always wrong. 

**To Reproduce**
Place ATV in different subnet. (I guess)
**Expected behavior**

**System Setup (please complete the following information):**
 - OS: Debian
 - Python: 3.7.3
 - pyatv: 0.7.0
- Apple TV: Apple TV 4K tvOS 13.4.6

**Additional context**
Traceback (most recent call last):
  File "/home/homeassistant/.homeassistant/custom_components/apple_tv/config_flow.py", line 144, in async_step_user
    await self.async_find_device()
  File "/home/homeassistant/.homeassistant/custom_components/apple_tv/config_flow.py", line 218, in async_find_device
    self.unique_id, self.hass.loop, cache=self.scan_result
  File "/home/homeassistant/.homeassistant/custom_components/apple_tv/config_flow.py", line 76, in device_scan
    scan_result = await scan(loop, hosts=hosts, timeout=timeout)
  File "/srv/homeassistant/lib/python3.7/site-packages/pyatv/__init__.py", line 49, in scan
    devices = (await scanner.discover(timeout)).values()
  File "/srv/homeassistant/lib/python3.7/site-packages/pyatv/support/scan.py", line 154, in discover
    *[self._get_services(host, timeout) for host in self.hosts]
  File "/srv/homeassistant/lib/python3.7/site-packages/pyatv/support/scan.py", line 167, in _get_services
    self.loop, str(host), ALL_SERVICES, port=port, timeout=timeout
  File "/srv/homeassistant/lib/python3.7/site-packages/pyatv/support/mdns.py", line 448, in unicast
    f"address {address} is not in any local subnet"
pyatv.exceptions.NonLocalSubnetError: address 192.168.178.49 is not in any local subnet

postlund:
This might be a bug with the new scanning implementation, I have seen some issues myself that I will look in to. Adding via unicast (specifying IP address) will not work between VLANs. That's just how zeroconf is designed.

[...]

PR #899, opened 2020-12-19
Improve DNS message parsing for mDNS

paxswill:
I also encountered the issues in #878/#765, so decided to take a crack at it. To make it easier to keep track of the return types and arguments, I replaced the data+ptr arguments and tuple return types with giving each function a BytesIO stream and using the position in that stream as the pointer.

As far as standards compliance goes, TXT record parsing was improved so that records longer than 63 bytes could be handled (I think that might've been contributing to #878) and the property keys are now compared case-insensitively to match the RFC better. I also changed some of the encode/decode calls with IDNA (for domain names) and ASCII (as DNS-SD keys are explicitly ASCII only).

I changed `Dns[Header|Question|Resource]` from `collections.namedtuple` to `typing.NamedTuple`, but I'm not completely sure I'm satisfied with them (maybe dataclasses? but that would need the dataclass backport for 3.6. Or an actual class hierarchy for different resource record types?)

With these changes, I'm now able to scan and discover my Apple TV that is on a different subnet (through avahi-reflector), as well as pair with the Apple TV. Sorry about the noise in other files from changing the query type constants into an `IntEnum`.


postlund:
I have honestly not looked at this yet, I will try to take time later tonight to do so (and test it). But I'm *very* grateful that you have taken time and tried to fix this mess. Truly appreciated!

postlund:
You seem to be on a roll. Just ping me when you are done 😅

pyatv/support/mdns.py
```diff
@@ -22,36 +22,41 @@
 logging.addLevelName(TRAFFIC_LEVEL, "Traffic")
 
 
-class QueryType(enum.IntEnum):
```
paxswill: The diff for this is not great, but `QueryType` and the DNS named tuples were moved further down in the file.

paxswill:
Ugh, those macOS failures. It looks like it's a known issue (pytest-dev/pytest-timeout#8), with a [behavior change in 3.8](https://bugs.python.org/issue38904) making it more visible. I can at least occasionally reproduce it locally, so I'll be poking at this for a bit.

Besides a fix for that (which I suspect is going to be to the testing setup, and probably not in this PR), I'm not planning on making other changes. I tried to keep the commits self-contained, but it does lead to some places being changed a few times (ex: parsing changes, with a later split into different modules).

postlund:
I'm gonna try to review this ASAP, but since it's huge and I only have my phone right now, it will probably take a while 😅

pyatv/support/dns.py
```diff
@@ -0,0 +1,344 @@
...
+        encoded_length = len(encoded_label)
+        # Length of the encoded label, in bytes, but a maximum of 63
+        # The maximum is 63 as the upper two bits are used as a flag for name
+        # compression.
+        encoded.append(min(encoded_length, 63))
+        if encoded_length == 0:
+            # If we've reached an empty label, assume this is the last component.
+            # Empty labels (two periods right after each other) isn't legal anyways.
+            break
+        # TODO: Should an error be raised if a label is too long?
```
postlund, line 39: I guess we should not end up here, but maybe log a warning or at least debug about it?

postlund (changes requested):
Doing such a big review thoroughly is really hard, but I've done my best to grasp what you have done and it's a big lift of my mess. So big thank you for that! 😊 Just noted two very minor things otherwise ready to go.

Just gonna throw it out there since you obviously have some domain knowledge... There's another bug (#765) where multicast requests are not reaching other networks when using Protocol-Independent Multicast. Or responses are not revived at least. If that's something you might have any ideas for troubleshooting, I would be very very interested...

pyatv/support/dns.py
```diff
@@ -0,0 +1,344 @@
...
+from ipaddress import IPv4Address
+
+from pyatv.support.collections import CaseInsensitiveDict
+
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def unpack_stream(fmt: str, buffer: typing.BinaryIO) -> typing.Tuple:
+    """Unpack a tuple from a binary stream, much like `struct.unpack`."""
```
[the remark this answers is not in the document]
paxswill: I rewrote this docstring to be closer to [`struct.unpack`'s](https://docs.python.org/3/library/struct.html#struct.unpack) (which uses "Unpack from *thing*, according to the format string")

pyatv/support/dns.py
paxswill: Yeah, I wasn't sure if raising a `ValueError` would be more appropriate than just logging it. If this was part of a more DNS focused-library maybe, but this is just a supporting module for Apple TV support. Speaking of which, I think I remember seeing macOS truncate names when they got too long too (but I haven't had such a long hostname in a while).

paxswill:
> Doing such a big review thoroughly is really hard, but I've done my best to grasp what you have done and it's a big lift of my mess. So big thank you for that! Just noted two very minor things otherwise ready to go.

Yeah, sorry not splitting it up a bit more :sweat_smile:

 In addition to those changes, I also made a small change to `DnsMessage.pack()` to use the `DnsHeader` and `DnsQuestion` `pack()` methods (I thought I'd made that change already :/ ).

> Just gonna throw it out there since you obviously have some domain knowledge... There's another bug (# 765) where multicast requests are not reaching other networks when using Protocol-Independent Multicast. Or responses are not revived at least. If that's something you might have any ideas for troubleshooting, I would be very very interested...

I haven't used Protocol-Independent Multicast before, but I might be able to get a simple setup going to reproduce it after another project I'm poking at.

postlund:
Looks good to me! 👍

postlund (approved):

PR #927, opened 2021-01-12
Use UTF-8 for DNS names, and (attempt) to handle dots in names

paxswill:
Unfortunately I was unable to add emoji to my Apple TV's name, but if Apple ever fixes that, this should handle that too.

### DIFF
--- a/pyatv/support/dns.py
+++ b/pyatv/support/dns.py
@@ -1,10 +1,12 @@
 """Processing functions for raw DNS messages."""
+import collections.abc
 import enum
 import io
 import logging
 import struct
 import typing
 from ipaddress import IPv4Address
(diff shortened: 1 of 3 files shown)


paxswill:
My old nemesis...

Just thought of a case this won't handle: if someone deliberately sets their Apple TV's name to include "_foo._tcp", meaning full names like `_foo._tcp._airplay._tcp.local` won't work (and fixing that would be a much larger diff).

postlund:
Tested on my end and my Apple TV 3 is now found again, so it seems to work 👍 Looks good otherwise, so I'll merge once update is complete. The case you mention sound a lot like a very obscure corner case, so I suggest that we skip it.

PR #968, opened 2021-02-16
protobuf: Bump version to 3.14.0

postlund:


postlund:
Deprecated because of #991.

PR #1009, opened 2021-04-12
Improve handling of unreliable connections

### LINKED ISSUE #1007: Multiple issues in MRP implementation triggered by unreliable connectivity in sleep mode
**Describe the bug**

When an Apple TV goes to sleep on a wired connection, connectivity may become unreliable due to an issue with the NIC and/or NIC driver on the Apple TV, leading to inefficient behavior in pyatv. (See https://discussions.apple.com/thread/8150492 for information on the NIC issue.)

While troubleshooting high resource utilization in home-assistant/core#49019, I discovered multiple issues in the connection handling portion of the MRP implementation:

- The existing heartbeat loop isn't ```cancel()```ed before starting a new connection and new heartbeat loop
- After ```stop()```ing, pending heartbeat messages will just ```start()``` a new connection (via ```send_and_receive()``` and ```_connect_and_encrypt()```)
- ```start()``` doesn't check if a connection attempt is in progress, resulting in multiple simultaneous connections attempts occasionally

These lead to an unnecessary spike in activity every time the Apple TV disconnects, as multiple heartbeats are sent, multiple new connections are attempted, etc, when the instance is – at least in Home Assistant – destined for destruction at that point.

**To Reproduce**

Put Apple TV to sleep on a wired connection.

**Expected behavior**

Resource utilization remains consistent.

**System Setup (please complete the following information):**
 - OS: Home Assistant
 - Python: 3.8.7
 - pyatv: 0.7.7
- Apple TV: Apple TV 4K

**Additional context**

doug-hoffman:
Resolves #1007

- Cleans up old connection by calling ```stop()``` before establishing a new connection
- Prevents unintentionally ```start()```ing a new connection after ```stop()```ing
- Prevents multiple parallel ```connect()``` calls

### DIFF
--- a/pyatv/mrp/protocol.py
+++ b/pyatv/mrp/protocol.py
@@ -5,6 +5,7 @@ import uuid
 import logging
 
 from collections import namedtuple
+from enum import Enum
 
 from pyatv import exceptions
 from pyatv.mrp import messages, protobuf
@@ -18,6 +19,25 @@ HEARTBEAT_RETRIES = 1  # One regular attempt + retries
 
 Listener = namedtuple("Listener", "func data")
 OutstandingMessage = namedtuple("OutstandingMessage", "semaphore response")
+
+
+class ProtocolState(Enum):
+    """Protocol internal state."""
+
+    NOT_CONNECTED = 0
+    """Protocol state is not connected."""
+
+    CONNECTING = 1
+    """Protocol state is connecting."""
+
+    CONNECTED = 2
+    """Protocol state is connected."""
+
+    READY = 3
+    """Protocol state is ready."""
+
+    STOPPED = 4
+    """Protocol state is stopped."""
 
 
 async def heartbeat_loop(protocol):
@@ -75,7 +95,7 @@ class MrpProtocol:
         self._heartbeat_task = None
         self._outstanding = {}
         self._listeners = {}
-        self._initial_message_sent = False
+        self._state = ProtocolState.NOT_CONNECTED
 
     def add_listener(self, listener, message_type, data=None):
         """Add a listener that will receice incoming messages."""
@@ -86,39 +106,52 @@ class MrpProtocol:
 
     async def start(self, skip_initial_messages=False):
         """Connect to device and listen to incoming messages."""
-        if self.connection.connected:
-            return
+        if self._state != ProtocolState.NOT_CONNECTED:
+            raise exceptions.InvalidStateError(self._state.name)
 
-        await self.connection.connect()
+        self._state = ProtocolState.CONNECTING
 
-        # In case credentials have been given externally (i.e. not by pairing
-        # with a device), then use that client id
-        if self.service.credentials:
-            self.srp.pairing_id = Credentials.parse(self.service.credentials).client_id
+        try:
+            await self.connection.connect()
 
-        # The first message must always be DEVICE_INFORMATION, otherwise the
-        # device will not respond with anything
-        msg = messages.device_information("pyatv", self.srp.pairing_id.decode())
+            self._state = ProtocolState.CONNECTED
 
-        self.device_info = await self.send_and_receive(msg)
-        self._initial_message_sent = True
+            # In case credentials have been given externally (i.e. not by pairing
+            # with a device), then use that client id
+            if self.service.credentials:
+                self.srp.pairing_id = Credentials.parse(
+                    self.service.credentials
+                ).client_id
 
-        # This is a hack to support re-use of a protocol object in
-        # proxy (will be removed/refactored later)
-        if skip_initial_messages:
-            return
+            # The first message must always be DEVICE_INFORMATION, otherwise the
+            # device will not respond with anything
+            msg = messages.device_information("pyatv", self.srp.pairing_id.decode())
 
-        await self._connect_and_encrypt()
+            self.device_info = await self.send_and_receive(msg)
 
-        # This should be the first message sent after encryption has
-        # been enabled
-        await self.send(messages.set_connection_state())
+            # This is a hack to support re-use of a protocol object in
+            # proxy (will be removed/refactored later)
+            if skip_initial_messages:
+                return
 
-        # Subscribe to updates at this stage
-        await self.send_and_receive(messages.client_updates_config())
-        await self.send_and_receive(messages.get_keyboard_session())
+            await self._enable_encryption()
 
-        self._heartbeat_task = asyncio.ensure_future(heartbeat_loop(self))
+            # This should be the first message sent after encryption has
+            # been enabled
+            await self.send(messages.set_connection_state())
+
+            # Subscribe to updates at this stage
+            await self.send_and_receive(messages.client_updates_config())
+            await self.send_and_receive(messages.get_keyboard_session())
+
+            self._heartbeat_task = asyncio.ensure_future(heartbeat_loop(self))
+        except Exception:
+            # Something went wrong, let's do cleanup
+            self.stop()
+            raise
+        else:
+            # We're now ready
+            self._state = ProtocolState.READY
 
     def stop(self):
         """Disconnect from device."""
@@ -130,19 +163,14 @@ class MrpProtocol:
         if self._heartbeat_task is not None:
             self._heartbeat_task.cancel()
             self._heartbeat_task = None
-        self._initial_message_sent = False
         self._outstanding = {}
         self.connection.close()
+        self._state = ProtocolState.STOPPED
 
-    async def _connect_and_encrypt(self):
-        if not self.connection.connected:
-            await self.start()
-
+    async def _enable_encryption(self):
         # Encryption can be enabled whenever credentials are available but only
         # after DEVICE_INFORMATION has been sent
-        if self.service.credentials and self._initial_message_sent:
-            self._initial_message_sent = False
-
+        if self.service.credentials:
             # Verify credentials and generate keys
             credentials = Credentials.parse(self.service.credentials)
             pair_verifier = MrpPairingVerifier(self, self.srp, credentials)
@@ -156,12 +184,21 @@ class MrpProtocol:
 
     async def send(self, message):
         """Send a message and expect no response."""
-        await self._connect_and_encrypt()
+        if self._state not in [
+            ProtocolState.CONNECTED,
+            ProtocolState.READY,
+        ]:
+            raise exceptions.InvalidStateError(self._state.name)
+
         self.connection.send(message)
 
     async def send_and_receive(self, message, generate_identifier=True, timeout=5):
         """Send a message and wait for a response."""
-        await self._connect_and_encrypt()
+        if self._state not in [
+            ProtocolState.CONNECTED,
+            ProtocolState.READY,
+        ]:
+            raise exceptions.InvalidStateError(self._state.name)
 
         # Some messages will respond with the same identifier as used in the
         # corresponding request. Others will not and one example is the crypto
--- a/docs/api/pyatv/exceptions.html
+++ b/docs/api/pyatv/exceptions.html
@@ -39,6 +39,9 @@ link_group: api
 <h4><code><a title="pyatv.exceptions.InvalidDmapDataError" href="#pyatv.exceptions.InvalidDmapDataError">InvalidDmapDataError</a></code></h4>
 </li>
 <li>
+<h4><code><a title="pyatv.exceptions.InvalidStateError" href="#pyatv.exceptions.InvalidStateError">InvalidStateError</a></code></h4>
+</li>
+<li>
 <h4><code><a title="pyatv.exceptions.NoAsyncListenerError" href="#pyatv.exceptions.NoAsyncListenerError">NoAsyncListenerError</a></code></h4>
 </li>
 <li>
@@ -153,7 +156,11 @@ class NonLocalSubnetError(Exception):
     &#34;&#34;&#34;Thrown when address it not in any local subnet.
 
     DEPRECATED: Not used since 0.7.1. Will be removed in 0.9.0!
-    &#34;&#34;&#34;</code></pre>
+    &#34;&#34;&#34;
+
+
+class InvalidStateError(Exception):
+    &#34;&#34;&#34;Thrown when trying to perform an action not possible in the current state.&#34;&#34;&#34;</code></pre>
 </details>
 </section>
 <section>
@@ -291,6 +298,25 @@ class NonLocalSubnetError(Exception):
 </summary>
 <pre><code class="python">class InvalidDmapDataError(Exception):
     &#34;&#34;&#34;Thrown when invalid DMAP data is parsed.&#34;&#34;&#34;</code></pre>
+</details>
+<h3>Ancestors</h3>
+<ul class="hlist">
+<li>builtins.Exception</li>
+<li>builtins.BaseException</li>
+</ul>
+</dd>
+<dt id="pyatv.exceptions.InvalidStateError"><code class="flex name class">
+<span>class <span class="ident">InvalidStateError</span></span>
+<span>(</span><span>*args, **kwargs)</span>
+</code></dt>
+<dd>
+<section class="desc"><p>Thrown when trying to perform an action not possible in the current state.</p></section>
+<details class="source">
+<summary>
+<span>Expand source code</span>
+</summary>
+<pre><code class="python">class InvalidStateError(Exception):
+    &#34;&#34;&#34;Thrown when trying to perform an action not possible in the current state.&#34;&#34;&#34;</code></pre>
 </details>
 <h3>Ancestors</h3>
 <ul class="hlist">
--- a/pyatv/exceptions.py
+++ b/pyatv/exceptions.py
@@ -70,3 +70,7 @@ class NonLocalSubnetError(Exception):
 
     DEPRECATED: Not used since 0.7.1. Will be removed in 0.9.0!
     """
+
+
+class InvalidStateError(Exception):
+    """Thrown when trying to perform an action not possible in the current state."""


postlund:
Thanks for the contribution! I need to take a closer look at this before I can review it. I do see some potential issues, but some assumptions are not correct (which is also understandable as it's not possible to know).

An `MrpConnection` is not re-usable, despite some legacy parts of the code might suggest otherwise. If a connection fails or if the connection is lost for some reason, it must be discarded and a new connection created, I.e. you need to use `pyatv.connect` again. The internals should never try to "fix" the situation under-the-hood, only let the listener know and tear everything down.

As `pyatv.connect` never returns before a connection has been made, there shouldn't be any race conditions between (potential) tasks calling methods that would trigger an additional connect. There might be gaps for internal code or if listeners try to call commands while a connection being torn down I guess. That's something that should be guarded against. Do you have any other concrete examples where this is happening?

doug-hoffman:
I'm not trying to fix anything beyond what MrpProtocol already did – if not connected, then connect. Actually, the changes in this PR make it less likely it will reconnect.

The problem is, since this is running async, the connection disconnects, then home assistant tries to ```close()``` the MrpAppleTv instance, then it'll run another loop of the heartbeat, which will then make ```_connect_and_encrypt()``` call another ```start()``` to send the heartbeat, then the old heartbeat loop runs again and finds out it should stop, but it's too late – we already have a new connection and new heartbeat loop, and may have lost control of of the old loop (depending on the exact order of events). And, to top it off, we've gone through all the trouble of connecting again – possibly multiple times – even though Home Assistant is about to abandon the instance.

The way I reworked the connection logic, ```_connect_and_encrypt``` only gets to make ```start()``` connect until ```stop()``` is called (unless ```start()``` is called again by something else). This way heartbeats can't make it reconnect. I also make sure ```stop()``` is called unconditionally before starting a new connection, so we properly ```cancel()``` the heartbeat loop and do any other necessary cleanup.

```start()```ing after ```stop()```ing happens almost every time it disconnects:

```
2021-04-12 21:09:52 DEBUG (MainThread) [pyatv.mrp.protocol] Starting heartbeat loop
2021-04-12 21:10:22 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 0
2021-04-12 21:10:22 DEBUG (MainThread) [pyatv.mrp.protocol] Got heartbeat 0
2021-04-12 21:10:52 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 1
2021-04-12 21:10:52 DEBUG (MainThread) [pyatv.mrp.protocol] Got heartbeat 1
2021-04-12 21:11:22 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 2
2021-04-12 21:11:23 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47472<->apple-tv:49153  Received EOF from server
2021-04-12 21:11:23 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47472<->apple-tv:49153  Disconnected from device: None
2021-04-12 21:11:23 WARNING (MainThread) [pyatv.mrp.protocol] There were 1 outstanding requests
2021-04-12 21:11:23 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47472<->apple-tv:49153  Closing connection
2021-04-12 21:11:23 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47472<->apple-tv:49153  Closing connection
2021-04-12 21:11:23 DEBUG (MainThread) [homeassistant.components.apple_tv] Starting connect loop
2021-04-12 21:11:23 DEBUG (MainThread) [homeassistant.components.apple_tv] Discovering device ID
2021-04-12 21:11:23 DEBUG (MainThread) [pyatv.mrp.protocol] heartbeat 2 failed
2021-04-12 21:11:23 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 3
2021-04-12 21:11:23 DEBUG (MainThread) [pyatv.mrp.protocol] Ignoring start while stopped
2021-04-12 21:11:23 ERROR (MainThread) [pyatv.mrp.protocol] heartbeat 3 failed after 2 tries
2021-04-12 21:11:23 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47472<->apple-tv:49153  Closing connection
2021-04-12 21:11:23 DEBUG (MainThread) [pyatv.mrp.protocol] Stopping heartbeat loop at 4
2021-04-12 21:11:25 DEBUG (MainThread) [homeassistant.components.apple_tv] Connecting to device Apple TV
2021-04-12 21:11:25 DEBUG (MainThread) [pyatv.mrp.connection]  Closing connection
2021-04-12 21:11:25 DEBUG (MainThread) [pyatv.mrp.protocol] Starting connection
2021-04-12 21:11:25 DEBUG (MainThread) [pyatv.mrp.connection] Connection made to device
2021-04-12 21:11:25 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47484<->apple-tv:49153  Connection established
2021-04-12 21:11:25 DEBUG (MainThread) [homeassistant.components.apple_tv] Changing address to apple-tv
2021-04-12 21:11:25 DEBUG (MainThread) [homeassistant.components.apple_tv] Connect loop ended
2021-04-12 21:11:25 DEBUG (MainThread) [pyatv.mrp.protocol] Starting heartbeat loop
2021-04-12 21:11:55 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 0
2021-04-12 21:12:00 DEBUG (MainThread) [pyatv.mrp.protocol] heartbeat 0 failed
2021-04-12 21:12:00 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 1
2021-04-12 21:12:03 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47484<->apple-tv:49153  Received EOF from server
2021-04-12 21:12:03 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47484<->apple-tv:49153  Disconnected from device: None
2021-04-12 21:12:03 WARNING (MainThread) [pyatv.mrp.protocol] There were 1 outstanding requests
2021-04-12 21:12:03 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47484<->apple-tv:49153  Closing connection
2021-04-12 21:12:03 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47484<->apple-tv:49153  Closing connection
2021-04-12 21:12:03 DEBUG (MainThread) [homeassistant.components.apple_tv] Starting connect loop
2021-04-12 21:12:03 DEBUG (MainThread) [homeassistant.components.apple_tv] Discovering device ID
2021-04-12 21:12:03 ERROR (MainThread) [pyatv.mrp.protocol] heartbeat 1 failed after 2 tries
2021-04-12 21:12:03 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47484<->apple-tv:49153  Closing connection
2021-04-12 21:12:03 DEBUG (MainThread) [pyatv.mrp.protocol] Stopping heartbeat loop at 2
```

If I change the new code to only warn and not interfere with the previous behavior, you can see how many connection attempts it would prevent:

```
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47324<->apple-tv:49153  Closing connection
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47324<->apple-tv:49153  Closing connection
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 3
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] WARN: start while stopped
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] Starting connection
2021-04-12 21:01:19 DEBUG (MainThread) [homeassistant.components.apple_tv] Starting connect loop
2021-04-12 21:01:19 DEBUG (MainThread) [homeassistant.components.apple_tv] Discovering device ID
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] heartbeat 0 failed
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 1
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] WARN: start while stopped
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] WARN: pending connection
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.support.knock] Knocking at ports [3689, 7000, 49152, 32498] on apple-tv
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] heartbeat 0 failed
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 1
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] WARN: start while stopped
2021-04-12 21:01:19 DEBUG (MainThread) [pyatv.mrp.protocol] Starting connection
```

And here's multiple heartbeat loops running before my change: (I only have a single Apple TV device)

```
2021-04-12 21:01:58 DEBUG (MainThread) [pyatv.mrp.protocol] Starting heartbeat loop
2021-04-12 21:02:00 DEBUG (MainThread) [pyatv.mrp.protocol] heartbeat 5 failed
2021-04-12 21:02:00 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 6
2021-04-12 21:02:00 DEBUG (MainThread) [pyatv.mrp.protocol] Got heartbeat 6
2021-04-12 21:02:00 ERROR (MainThread) [pyatv.mrp.protocol] heartbeat 3 failed after 2 tries
2021-04-12 21:02:00 DEBUG (MainThread) [pyatv.mrp.connection] home-assistant:47328<->apple-tv:49153  Closing connection
2021-04-12 21:02:00 DEBUG (MainThread) [pyatv.mrp.protocol] Stopping heartbeat loop at 4
2021-04-12 21:02:25 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 0
2021-04-12 21:02:25 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 2
2021-04-12 21:02:28 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 0
2021-04-12 21:02:30 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 7
2021-04-12 21:02:30 DEBUG (MainThread) [pyatv.mrp.protocol] heartbeat 0 failed
2021-04-12 21:02:30 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 1
2021-04-12 21:02:30 DEBUG (MainThread) [pyatv.mrp.protocol] heartbeat 2 failed
2021-04-12 21:02:30 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 3
2021-04-12 21:02:33 DEBUG (MainThread) [pyatv.mrp.protocol] heartbeat 0 failed
2021-04-12 21:02:33 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 1
2021-04-12 21:02:35 DEBUG (MainThread) [pyatv.mrp.protocol] heartbeat 7 failed
2021-04-12 21:02:35 DEBUG (MainThread) [pyatv.mrp.protocol] Sending periodic heartbeat 8
```

My solution may not be the best – but I'm confident that there's a problem that needs fixed here. Let me know if there's anything else in need of clarification.

**Comment by postlund** [comment]
Ok, I can see the problem here (pretty much as I suspected, but wasn't totally sure). I however think we can simplify the implementation a bit. Here are a few things we know:

* Since we never return a handler to an instance of `interface.AppleTV` before being connected, we can always (with one exception, explained next) expect us to be connected in `_connect_and_encrypt`
* The `start` method is called from `pyatv.mrp.AppleTV.connect`, so we also know that that method has been called
* It also not allowed to call `start` more than once
* An instance of `MrpConnection` can and shall never be re-used for any purpose
* Trying to send something once the underlying connection has been closed is an invalid action and should preferably result in an exception

So, I suggest something along these lines:

* Rename `_connect_and_encrypt` to `_enable_encryption` as that is what it's actually doing (and remove the check in the beginning, that potentially calls `start`)
* Add a flag, e.g. `_is_connected` that is set to `True` when `start` has been called (at the end, when we know that we have been connected)
* If `start` is called and this flag is `True`, raise an exception (maybe add a new one, e.g. `InvalidStateError`?)
* Do the same thing in `send` and `send_and_receive` as well but check if `False` instead (i.e. raise if not connected), effectively replacing the calls to `_enable_encryption` which should only be called from `start` now
* Set the flag to `False` in `stop`

This is the simple way, it's a bit fragile but works as long as everything is done properly. A more correct way would be to introduce an internal state, being one of `NotConnected`, `Connecting` or `Connected`. We would initialize to `NotConnected`, move to `Connecting` when `start` has been called and finally move on to  `Connected` at the end of `start`. We would also have to handle any error in `start` and move to `NotConnected`. After that, `start` can only be called when being in `NotConnected` state and `send`/`send_and_receive` when being in `Connected`.

**Comment by doug-hoffman** [comment]
Thoughts?

Using a boolean seemed a little problematic, since it misses multiple parallel attempts to connect before any one is able to complete, so I went with tracking the internal state. I had to add a 4th state ("Ready") to the list, since ```send()``` and ```send_and_receive()``` had to be able to operate before ```start()``` was fully complete. It seemed like a good idea to indicate when ```start()``` actually finished, in case it becomes important down the road.

This also had a similar effect on CPU utilization as what I had done before:

image

File: pyatv/const.py [hunk 1]
```diff
@@ -286,3 +286,19 @@ class FeatureName(Enum):
 
     TurnOff = 34
     """Turn off device."""
+
+
+class ProtocolState(Enum):
```
postlund, line 291: This file (`const.py`) is only for things in the public interface, so you should move this back to `protocol.py` as the protocol state is only internal.

pyatv/mrp/protocol.py
```diff
@@ -86,39 +88,53 @@ def add_listener(self, listener, message_type, data=None):
 
     async def start(self, skip_initial_messages=False):
         """Connect to device and listen to incoming messages."""
-        if self.connection.connected:
-            return
+        if self._state != ProtocolState.NotConnected:
+            raise exceptions.InvalidStateError(ProtocolState(self._state).name)
```
postlund, line 92:
```suggestion
            raise exceptions.InvalidStateError(self._state.name)
```

pyatv/mrp/protocol.py
```diff
@@ -133,11 +149,9 @@ def stop(self):
         self._initial_message_sent = False
         self._outstanding = {}
         self.connection.close()
+        self._state = ProtocolState.NotConnected
```
postlund, line 152: This is the logical state here, but I think I would like to introduce another state (e.g. `Stopped`) as a way to prevent re-use of the protocol instance. Because now you can call `start` again after `stop` and I don't see any point of doing that, so it's better to block in order to find potential bugs.

pyatv/mrp/protocol.py
```diff
@@ -133,11 +149,9 @@ def stop(self):
...
 
-    async def _connect_and_encrypt(self):
-        if not self.connection.connected:
-            await self.start()
-
+    async def _enable_encryption(self):
         # Encryption can be enabled whenever credentials are available but only
         # after DEVICE_INFORMATION has been sent
         if self.service.credentials and self._initial_message_sent:
```
postlund, line 157: This is just a clean-up I saw now. Since we only call this method after sending the initial message, there's really no point of having a check for that (it will always be `True` here). So you can remove all instances of `self._initial_message_sent` in the file.

pyatv/mrp/protocol.py
```diff
@@ -156,12 +170,24 @@ async def _connect_and_encrypt(self):
 
     async def send(self, message):
         """Send a message and expect no response."""
-        await self._connect_and_encrypt()
+        if self._state not in [
+            ProtocolState.Connected,
+            ProtocolState.Ready,
+        ]:
+            raise exceptions.InvalidStateError(ProtocolState(self._state).name)
```
postlund, line 177:
```suggestion
            raise exceptions.InvalidStateError(self._state.name)
```

pyatv/mrp/protocol.py
```diff
@@ -156,12 +170,24 @@ async def _connect_and_encrypt(self):
...
         self.connection.send(message)
 
     async def send_and_receive(self, message, generate_identifier=True, timeout=5):
         """Send a message and wait for a response."""
-        await self._connect_and_encrypt()
+        if self._state not in [
+            ProtocolState.Connected,
+            ProtocolState.Ready,
+        ]:
+            raise exceptions.InvalidStateError(ProtocolState(self._state).name)
```
postlund, line 188:
```suggestion
            raise exceptions.InvalidStateError(self._state.name)
```

pyatv/mrp/protocol.py
```diff
@@ -156,12 +170,24 @@ async def _connect_and_encrypt(self):
...
     async def send(self, message):
         """Send a message and expect no response."""
-        await self._connect_and_encrypt()
+        if self._state not in [
+            ProtocolState.Connected,
+            ProtocolState.Ready,
+        ]:
+            raise exceptions.InvalidStateError(ProtocolState(self._state).name)
+
+        await self._enable_encryption()
```
postlund, line 179: Remove

pyatv/mrp/protocol.py
```diff
@@ -156,12 +170,24 @@ async def _connect_and_encrypt(self):
...
     async def send_and_receive(self, message, generate_identifier=True, timeout=5):
         """Send a message and wait for a response."""
-        await self._connect_and_encrypt()
+        if self._state not in [
+            ProtocolState.Connected,
+            ProtocolState.Ready,
+        ]:
+            raise exceptions.InvalidStateError(ProtocolState(self._state).name)
+
+        await self._enable_encryption()
```
postlund, line 190: Remove

postlund (changes requested):
I like the state based version better than a bool, so great that you picked that one 👍 A few minor comments to fix, after that we'll merge.

doug-hoffman:
That should address everything. Seems to be working well here. Let me know if there's anything else.

postlund (approved):
This looks good to me, lets merge! 👍

PR #1022, opened 2021-04-20
if: Add Apps interface to Companion

### LINKED ISSUE #303: Support opening an app (like with Siri shortcuts)
Since it's possible to open apps on tvOS using Siri shortcuts, it would be great if `pyatv` could do that as well. I am not sure how that API works or even which messages are involved, so that would be a first step to investigate. Adding a new interface to launch the app needs to be done.

An idea is to see if the proxy can be used to intercept the traffic, ("launch app on ATVProxy") but my tests have failed so far so it needs some additional work.

**What needs to be done?**
TBD

No

Somewhat relates to #302. A generic API should be added but only `MRP` will implement it.

### LINKED ISSUE #1020: Support listing apps installed on Apple TV
**What feature would you like?**
App launching is coming soon. Once #655 is done, I'll continue with #303 to finish that. Launching apps however require the bundle ID for the app to launch, e.g. `com.netflix.Netflix` for Netflix, which isn't very friendly to use. Since it's possible to fetch all installed apps (or at least "launchable apps") and get mapping between bundle ID and name that way, it would be great to add app list as a feature as well.

Worth noting is that app names are returned localized in the language configured on the Apple TV. Don't think I can do anything about that.

**Describe the solution you'd like**
A simple method that returns a list of `interface.App` should be sufficient.

**Any other information to share?**
Protocol related information can be found in #655. It is documented in the PR as well (#657).

postlund:
Both launching apps and fetching list of apps are supported.

Fixes #303
Fixes #1020

---

To try this out, both `MRP` and `Companion` credentials are needed. So start by pairing and saving them (unless you already have them previously stored):

```shell
$ atvremote -s <ip> --protocol mrp pair
$ atvremote -s <ip> --protocol companion pair
```

Now, to list apps:

```shell
$ atvremote -s <ip> --mrp-credentials <mrp credentials> --companion-credentials <companion credentials> app_list
App: Podcaster (com.apple.podcasts), App: Filmer (com.apple.TVMovies), App: TV (com.apple.TVWatchList), App: Bilder (com.apple.TVPhotos), App: App Store (com.apple.TVAppStore), App: C More (se.cmore.CMore2), App: Arcade (com.apple.Arcade), App: Sök (com.apple.TVSearch), App: Emby (emby.media.emby-tvos), App: TV4 Play (se.tv4.tv4play), App: Datorer (com.apple.TVHomeSharing), App: YouTube (com.google.ios.youtube), App: SVT Play (se.svtplay.mobil), App: Plex (com.plexapp.plex), App: Viafree (com.MTGx.ViaFree.se), App: Inställningar (com.apple.TVSettings), App: Apple Events (com.apple.appleevents), App: discovery+ (com.kanal5.play), App: Netflix (com.netflix.Netflix), App: Viaplay (se.harbourfront.viasatondemand), App: Musik (com.apple.TVMusic)
```

To launch an app, pass the bundle id to `launch_app`:
```
$ atvremote -s <ip> --mrp-credentials <mrp credentials> --companion-credentials <companion credentials> launch_app=com.netflix.Netflix
```
Hopefully it should launch/switch to the app OK. Unfortunately, there doesn't seem to be any feedback if it did or not. So you won't get any output from `atvremote` regarding that.


nehalvpatel:
I tested this and it worked flawlessly. It listed every single app I have. I made a little script to launch every single app from that list of bundle ids, and it opened all of them. ✅

To people wondering how to get this, install it using this command. I had to delete my existing virtual environment and create a new one before it recognized the new version:
```
pip3 install git+https://github.com/postlund/pyatv.git@refs/pull/1022/head
```

My **only suggestion** concerning this would be to alphabetize the list returned by the `app_list` command.

nehalvpatel:
Screen Shot 2021-04-20 at 9 39 55 PM

I got this error while testing on my Raspberry Pi 4. I also got it after using the pair command for both the protocols, so I'm not sure if it's related to this pull request.

Oddly enough, the commands still go through despite this error. I'm able to pair using both protocols and launch apps even though this error kept showing up.

postlund:
@nehalvpatel Yeah, that is fine. The `knock` module is used to send packets to the device to wake it up in case it's sleeping. A background task is started when connecting and cancelled when the connection succeeds and ultimately, I should just catch that exception and suppress it. I will fix that today! Great to hear that it is working 👍

postlund:
Fixed the `knock` error in #1023, which is merged and rebased this PR on top of that. So the error should not appear anymore, I hope.

WizBangCrash:
I tested this using macOS 10.15.7 on an ATV4K running tvOS 14.4 and worked flawlessly :-)

postlund:
@WizBangCrash Perfect, thanks for testing! 😄

postlund:
Merged 🎉

pdecrom:
Hi, I also tried the new feature app_list and app_launch. The app_launch works fine with latest master and I can e.g. start the apps with bundles:
- com.amazon.aiv.AIVApp
- com.netflix.Netflix

but getting the app_list crashes:

[debug.log](https://github.com/postlund/pyatv/files/6366758/debug.log)

postlund:
@pdecrom Interesting. Can you list apps using Shortcuts? Did you also try restarting the Apple TV (try Shortcuts before doing that though)? Just want to pinpoint if it's something in `pyatv` or related to the Apple TV in general, sometimes it is.

pdecrom:
Hi, yes I just tried the shortcuts on my iphone. When adding a shortcut to open an app on the appletv, all my apps are listed.
The apple tv was already restarted after I first encountered the issue.

Hey, and guess what, now that I have used those shortcuts for the first time from my iphone, I get all apps back from app_list.

postlund:
Hmm, that's a peculiar case. The error message you got, `No request handler`, seems to mean that nobody has registered at the receiving end to handle that particular command (`FetchLaunchableApplicationsEvent`). Maybe the Shortcuts app "start something" (because of a more proper set up) that manages this command and that's why it works afterwards. Does it still work even if you restart the device and not try to fetch app list in Shortcuts before using `app_list`?

pdecrom:
I think you are on the right track, after restarting the ATV, the app_list command crashes with the same error.
But right after creating a shortcut and listing the apps on the iPhone, the app_list command works.
Looks like something is still missing to init that handler correctly.

postlund:
Definitely! Thanks for confirming this. Would you mind creating a bug issue, include log and findings there so it can be tracked?

PR #1064, opened 2021-05-10
some small additions to the OPACK definitions

jdsnape:
<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1064"></a>


jdsnape:
> Looks good to me  Did you encounter any of the new ones in the wild?

I've only seen the 0xC1 item itw -> the homepod has a "_roomName" key in the systeminfo response which I think is an ID (I'm not entirely clear how that is mapped to the name in the home app - still todo!)

PR #1135, opened 2021-06-12
raop: Add support for volume controls

### LINKED ISSUE #1129: Support for the audio interface in RAOP
I added the audio interface a while back in #1074, but it has not been implemented for any protocol yet. It should be pretty easy to implement for RAOP, so I would like to do that before releasing 0.8.0. I should also implement `volume_up` and `volume_down` in the remote control interface.

**What needs to be done?**

* Read out current volume and don't set a default volume level (I think this is possible by just sending `volume: 0.0` and I also think it's available via `/info` or `/server-info` for devices supporting those endpoints)
* Implement the audio interface, which should be trivial
* Add the remote control interface and implement volume controls
* Update documentation and tests

No

When I experimented with the DACP remote, e.g. getting events about buttons being pressed on the receiver, I did actually get volume updates. That was the only thing I got, so I should probably implement support for those events as well to be more complete.

postlund:
Fixes #1129

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1135"></a>

### DIFF
--- a/docs/api/pyatv/interface.html
+++ b/docs/api/pyatv/interface.html
@@ -3472,6 +3472,7 @@ def stop(self) -&gt; None:
 <ul class="hlist">
 <li>pyatv.dmap.DmapRemoteControl</li>
 <li>pyatv.mrp.MrpRemoteControl</li>
+<li>pyatv.raop.RaopRemoteControl</li>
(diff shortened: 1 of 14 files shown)


postlund:
Latest version of bitarray breaks all PRs right now, so I need to restrict supported versions and exclude latest version.

PR #1208, opened 2021-07-13
Update README.md

xpnewmedia:
fixed getting started link on README.md.
I always come here to actually get started with the library,
almost everytime.

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1208"></a>


postlund (approved):

postlund:
Ah, right, that is correct. Thank you! 👍

Issue #556, opened 2020-04-02
AirPlay does not support user specified password


postlund:
**Describe the bug**
If AirPlay is configured with user provided password, pairing will work as expected but playing something with `play_url` will not. This is because `pyatv` does not support this kind of authentication.

**To Reproduce**
* Enable password on Apple TV
* Pair with AirPlay and get credentials
* Try to play something with `play_url`
* Get 401 error

**Expected behavior**
Pairing and playing should work

**System Setup (please complete the following information):**
 - OS: any
 - Python: any
 - pyatv: any
- Apple TV: any

**Additional context**
It is described here: https://nto.github.io/AirPlay.html#passwordprotection

TkachenkoViacheslav:
Hi, @postlund 
Do you have any updates or announcements about AirPlay password feature implementation?

PR #1212, opened 2021-07-14
[RAOP] Add support for password protected devices

Schlaubischlump:
This pull request implements support for password protected RAOP devices. I'm ready to implement any changes / cleanups that are required.

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1212"></a>

### DIFF
--- a/pyatv/raop/__init__.py
+++ b/pyatv/raop/__init__.py
@@ -290,6 +290,7 @@ class RaopStream(Stream):
         try:
             client, _, context = await self.playback_manager.setup()
             client.credentials = self._get_credentials()
+            client.password = self.service.password
 
             client.listener = self.listener
             await client.initialize(self.service.properties)
--- a/pyatv/raop/rtsp.py
+++ b/pyatv/raop/rtsp.py
@@ -1,9 +1,10 @@
 """Implementation of the RTSP protocol."""
 import asyncio
+from hashlib import md5
 import logging
 import plistlib
 from random import randrange
-from typing import Dict, Mapping, Optional, Tuple, Union
+from typing import Dict, Mapping, NamedTuple, Optional, Tuple, Union
 
 from pyatv.dmap import tags
 from pyatv.raop import timing
@@ -42,6 +43,28 @@ CURVE25519_PUB_KEY = (
 )
 
 
+class DigestInfo(NamedTuple):
+    """
+    OAuth information used for password protected devices.
+    """
+
+    username: str
+    realm: str
+    password: str
+    nonce: str
+
+
+def get_digest_payload(method, uri, user, realm, pwd, nonce):
+    """Return the Authorization payload for Apples OAuth."""
+    payload = (
+        'Digest username="{0}", realm="{1}", nonce="{2}", uri="{3}", response="{4}"'
+    )
+    ha1 = md5(f"{user}:{realm}:{pwd}".encode("utf-8")).hexdigest()
+    ha2 = md5(f"{method}:{uri}".encode("utf-8")).hexdigest()
+    di_response = md5(f"{ha1}:{nonce}:{ha2}".encode("utf-8")).hexdigest()
+    return payload.format(user, realm, nonce, uri, di_response)
+
+
 class RtspContext:
     """Data used for one RTSP session.
 
@@ -69,6 +92,9 @@ class RtspContext:
         self.session_id: int = randrange(2 ** 32)
         self.dacp_id: str = f"{randrange(2 ** 64):X}"
         self.active_remote: int = randrange(2 ** 32)
+
+        # Required for password authentication
+        self.digest_info: Optional[DigestInfo] = None
 
         self.volume: Optional[float] = None
 
@@ -143,7 +169,7 @@ class RtspSession:
             body=body,
         )
 
-    async def announce(self) -> HttpResponse:
+    async def announce(self, password: Optional[str]) -> HttpResponse:
         """Send ANNOUNCE message."""
         body = ANNOUNCE_PAYLOAD.format(
             session_id=self.context.session_id,
@@ -153,11 +179,30 @@ class RtspSession:
             channels=self.context.channels,
             sample_rate=self.context.sample_rate,
         )
-        return await self.exchange(
+
+        requires_password: bool = password is not None
+
+        response = await self.exchange(
             "ANNOUNCE",
             content_type="application/sdp",
             body=body,
+            allow_error=requires_password,
         )
+
+        # Save the necessary data for password authentication
+        www_authenticate = response.headers.get("www-authenticate", None)
+        if response.code == 401 and www_authenticate and requires_password:
+            _, realm, _, nonce, _ = www_authenticate.split('"')
+            info = DigestInfo("pyatv", realm, password, nonce)  # type: ignore
+            self.context.digest_info = info
+
+            response = await self.exchange(
+                "ANNOUNCE",
+                content_type="application/sdp",
+                body=body,
+            )
+
+        return response
 
     async def setup(self, control_port: int, timing_port: int) -> HttpResponse:
         """Send SETUP message."""
@@ -242,6 +287,13 @@ class RtspSession:
             "Active-Remote": self.context.active_remote,
             "Client-Instance": self.context.dacp_id,
         }
+
+        # Add the password authentication if required
+        if self.context.digest_info:
+            hdrs["Authorization"] = get_digest_payload(
+                method, uri or self.uri, *self.context.digest_info
+            )
+
         if headers:
             hdrs.update(headers)
 
--- a/pyatv/scripts/atvremote.py
+++ b/pyatv/scripts/atvremote.py
@@ -437,6 +437,13 @@ async def cli_handler(loop):
         default=False,
     )
 
+    parser.add_argument(
+        "--raop-password",
+        help="optional password for raop",
+        dest="raop_password",
+        default=None,
+    )
+
     creds = parser.add_argument_group("credentials")
     for prot in Protocol:
         creds.add_argument(
@@ -527,6 +534,10 @@ async def _autodiscover_device(args, loop):
     for proto in Protocol:
         _set_credentials(proto, f"{proto.name.lower()}_credentials")
 
+    raop_service = apple_tv.get_service(Protocol.RAOP)
+    if raop_service:
+        raop_service.password = args.raop_password
+
     logging.info("Auto-discovered %s at %s", apple_tv.name, apple_tv.address)
 
     return apple_tv
@@ -550,7 +561,12 @@ def _manual_device(args):
         )
     if args.raop_credentials or args.protocol == const.Protocol.RAOP:
         config.add_service(
-            RaopService(args.id, args.port, credentials=args.raop_credentials)
+            RaopService(
+                args.id,
+                args.port,
+                credentials=args.raop_credentials,
+                password=args.raop_password,
+            )
         )
     return config
 
--- a/tests/fake_device/raop.py
+++ b/tests/fake_device/raop.py
@@ -1,8 +1,11 @@
 """Fake RAOP device for tests."""
 import asyncio
 from functools import wraps
+from hashlib import md5
 import logging
 import plistlib
+import random
+import string
 from types import SimpleNamespace
 from typing import Dict, Optional, cast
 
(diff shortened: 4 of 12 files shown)


pyatv/scripts/atvremote.py
```diff
@@ -427,6 +427,12 @@ async def cli_handler(loop):
         default=False,
     )
 
+    parser.add_argument(
+        "--raop_password",
```
postlund, line 431: As mentioned before, just - instead of _ (just a reminder).

pyatv/raop/rtsp.py
```diff
@@ -153,12 +163,30 @@ async def announce(self) -> HttpResponse:
...
+            _, realm, _, nonce, _ = www_authenticate.split("\"")
+            self.context.digest_info = DigestInfo("pyatv", realm, password, nonce)
+
+            response = await self.exchange(
+                "ANNOUNCE",
+                content_type="application/sdp",
+                body=body,
+            )
+
+        return response
```
postlund, line 196: If I understand the implementation correctly, we will just fall-though here and rely on the next call raising an exception in case a password is required but none is provided? Maybe we can be helpful and raise a proper exception in that case (with a string mentioning a password is required)?

pyatv/support/facade.py
```diff
@@ -432,6 +432,9 @@ def service(self) -> interface.BaseService:
 
         raise RuntimeError("no service (bug)")
 
+    def get_service(self, protocol: Protocol) -> interface.BaseService:
```
postlund: This should not be here (it would also have to be added to `interface.AppleTV` if it were). See my comment in the test on what to do.

tests/raop/test_raop_functional.py
```diff
@@ -135,6 +136,20 @@ async def test_stream_complete_legacy_auth(
...
+@pytest.mark.parametrize(
+    "raop_properties,raop_server_password,raop_client_password",
+    [({"et": "0"}, "test", "test"), ({"et": "0"}, None, None)],
+)
+async def test_stream_with_password(
+    raop_client, raop_state, raop_usecase, raop_server_password, raop_client_password
+):
+    raop_usecase.password(raop_server_password)
+
+    raop_service = raop_client.get_service(Protocol.RAOP)
```
postlund, line 148: The `raop_conf` fixture holds the config, add that and get the service from there instead.

https://github.com/postlund/pyatv/blob/809c5a384cd7762410919d1e6cdc53f3e973f93f/tests/raop/conftest.py#L33

tests/fake_device/raop.py
```diff
@@ -451,6 +507,11 @@ def require_auth(self, is_required: bool) -> None:
         """Enable or disable requirement to perform authentication."""
         self.state.auth_required = is_required
 
+    def password(self, new_password: str) -> None:
```
postlund, line 510: Must be `Optional[str]` to support `None` as value.

tests/fake_device/raop.py
```diff
@@ -53,6 +100,9 @@ class FakeRaopState:
     def __init__(self):
         self.metadata = SimpleNamespace(title=None, artist=None, album=None)
         self.audio_packets: Dict[int, bytes] = {}  # seqo -> raw audio
+        self.password = None
```
postlund, line 103: Please add type hints to these.

tests/fake_device/raop.py
```diff
@@ -53,6 +100,9 @@ class FakeRaopState:
...
+        self.nonce = None
+        self.realm = "raop"
```
postlund, line 105: Maybe just put as a constant at the top?

tests/raop/test_raop_functional.py
```diff
@@ -135,6 +136,20 @@ async def test_stream_complete_legacy_auth(
     assert raop_state.auth_setup_performed == require_auth
     assert audio_matches(raop_state.raw_audio, frames=10)
 
+@pytest.mark.parametrize(
+    "raop_properties,raop_server_password,raop_client_password",
+    [({"et": "0"}, "test", "test"), ({"et": "0"}, None, None)],
+)
+async def test_stream_with_password(
```
postlund, line 143: I miss some kind of verification that password validation was performed. Maybe you can add a bool that you set after a correct password has been provided, more or less like `auth_setup_performed`? It's not great, but it at least allows for some additional confidence that everything worked.

postlund (changes requested):
Overall, looks good 👍 I added a few comments you can have a look at.

tests/fake_device/raop.py
Schlaubischlump: In theory realm can have different values. According to the unofficial Airplay documentation the Apple TV allows `AirPlay` and `raop`. Therefore I decided against a constant. I'm not quite sure, when which value is used.

tests/raop/test_raop_functional.py
Schlaubischlump: I think the nicer way would be to catch the `AuthenticationError`, since that is the expected behaviour. Sadly putting a try catch around the call doesn't seem to work. Does the `async` stuff somehow prevent the correct propagation of exceptions ?

tests/fake_device/raop.py
Schlaubischlump: Old python2 habits die hard. I will fix this.

pyatv/raop/rtsp.py
Schlaubischlump: I'm not sure I'm understanding you correctly. If the user does not provide a password and a password is required
`allow_error` will be `false`. That means the first `ANNOUNCE` will raise an exception for the `AuthenticationError`.

pyatv/raop/__init__.py
```diff
@@ -285,11 +285,13 @@ async def stream_file(self, file: Union[str, io.BufferedReader], **kwargs) -> No
 
         INCUBATING METHOD - MIGHT CHANGE IN THE FUTURE!
         """
+        audio_file = None
```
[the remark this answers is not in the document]
Schlaubischlump: Yes old stuff. Without this change a variable is not defined error instead of an `AuthenticationError` was thrown. I saw that you already added this change.

pyatv/raop/rtsp.py
postlund: Ah, yes, I missed that! You are correct, that is handled good enough (I.e. like before), so I'm satisfied with that 👍

tests/raop/test_raop_functional.py
postlund: Sure, you can do that as well. But there won't be any exception when the password is correct, so I guess you need separate test cases for positive contra negative outcome?

Shouldn't be any problem with async in regards to exceptions. There are a few layers in the cake we know as `stream_file`... Maybe it is swallowed somewhere along the way and not propagated back to the caller properly (by accident)?

Schlaubischlump: I don't know, what I made wrong the first time, but this time, I have no problem catching the exception this time... What do you think about the change ?

postlund: Yep, that works! 👍

postlund (changes requested):
I'm fine with code and docs so far 👍 It struck me though that there are two additional places in the docs that need changes:

* `development/stream.md`: First section states that password is not supported, just remove that. I would also like a short example of how to programmatically set a password under "Stream a file" so that someone using the library knows how to do that.
* `documentation/supported_features.md`: Add that password is supported under "Supported features" and remove that it isn't from "Limitations and notes"

After that we everything should be fine and dandy!

Schlaubischlump:
I only implemented password support for RAOP. I'm not sure how passwords are handled for airplay (maybe the same with `realm=Airplay`?). Therefore I have not removed the section completely.

postlund:
> I only implemented password support for RAOP. I'm not sure how passwords are handled for airplay (maybe the same with `realm=Airplay`?). Therefore I have not removed the section completely.
> 
> I have not implemented a setter in `conf.AppleTV` on purpose, because a setter such as `set_password` would suggest that it would also work for other protocols. A setter like `set_raop_password` must be removed in the future, if we add password support for other protocols. Maybe taking a look at the AirPlay protocol and how passwords are handled there is a good next step. Maybe we can merge this change first, and I'll take a look at the password stuff for the AirPlay protocol in the coming days / weeks. If I manage to support password there as well I'll add the setter.
> 
> If this approach is okay for you, I would add a disclaimer to the `development/stream.md` file, that the API is likely to change in the future.

I believe what you have right now is good for now. Adding similar support for password in AirPlay would be great as well, there's a very old issue for that already (#556). You are more than welcome to explore and implement that as well!

For now I don't see a need for a `set_password` method, so we can probably skip that. I'm not entirely satisfied with `set_credentials` to be honest. So let's keep it simple.

postlund:
Once checks pass I will merge this!

postlund:
@Schlaubischlump There's a few minor issue to fix. Clean those up and I'll merge ASAP.

Schlaubischlump:
Let me take a look.

postlund:
I think re-generating api docs is what's missing (that is only verified by python 3.8 now because of commit e1bc26f6010bf8ea4a598d9bbca0cdfb00df4be4.

Schlaubischlump:
Okay, I renamed the remaining password variable to get around the insecure crypto warning and executed the `api.py` script (using python3.8 just to be sure) to generate the documentation. Hopefully it is working now.

Schlaubischlump:
Strange... Why is it still failing ?

postlund:
> Strange... Why is it still failing ?

A few tests are a bit unstable (probably time sensitive). I haven't paid much attention to them in the past, just re-triggered and hoped that it worked next time... I'm gonna start writing them down now though and try to fix them.

postlund:
Uhm, GitHub complains about conflict when rebasing, maybe you can try that manually? Not sure where that came from, didn't complain about it before.

Schlaubischlump:
Do we have to fix the CodeQL error as well ? Because I'm out of ideas of to prevent it

postlund:
Nah, ignore CodeQL.

Schlaubischlump:
If I locally rebase I get a runtime error inside `daap.py` when I try to run the a simple test program. 

```
  File "/usr/local/lib/python3.9/site-packages/pyatv/scripts/atvremote.py", line 724, in _run_application
    return await cli_handler(loop)
  File "/usr/local/lib/python3.9/site-packages/pyatv/scripts/atvremote.py", line 504, in cli_handler
    return await _handle_commands(args, config, loop)
  File "/usr/local/lib/python3.9/site-packages/pyatv/scripts/atvremote.py", line 612, in _handle_commands
    atv = await connect(config, loop, protocol=args.protocol)
  File "/usr/local/lib/python3.9/site-packages/pyatv/__init__.py", line 127, in connect
    await atv.connect()
  File "/usr/local/lib/python3.9/site-packages/pyatv/support/facade.py", line 412, in connect
    await protocol_connect()
  File "/usr/local/lib/python3.9/site-packages/pyatv/dmap/__init__.py", line 606, in _connect
    await requester.login()
  File "/usr/local/lib/python3.9/site-packages/pyatv/dmap/daap.py", line 99, in login
    resp = await self._do(_login_request, is_login=True)
  File "/usr/local/lib/python3.9/site-packages/pyatv/dmap/daap.py", line 130, in _do
    resp, status = await action()
  File "/usr/local/lib/python3.9/site-packages/pyatv/dmap/daap.py", line 92, in _login_request
    url = self._mkurl("login?[AUTH]&hasFP=1", session=False, login_id=True)
  File "/usr/local/lib/python3.9/site-packages/pyatv/dmap/daap.py", line 157, in _mkurl
    if re.match(r"0x[0-9A-Fa-f]{16}", self._login_id):
  File "/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/re.py", line 191, in match
    return _compile(pattern, flags).match(string)
TypeError: expected string or bytes-like object
```

postlund:
You don't have to close the PR, just force-push your other branch (raop2) to this one and it will sort itself.

Schlaubischlump:
I wasn't sure how github handles a force push, but it seems to work just fine. Can you run the test again ?

postlund:
It works pretty ok fo force push (you need to do that if you don't use merge commits but rather rebase on top of master when catching up, like I prefer). But it can mess up open comments. Should be fine now though 👍

Restarted, let's hope it passes now!

postlund:
Finally merged, thanks for the job and patience here 👍🎉 Hopefully, future PRs won't be as troublesome.

Schlaubischlump:
Now that I know which linters etc. I have to run first, it should be easier the next time !

postlund:
Yeah, it takes some time to get used to. But once you do, it's pretty ok. A pre-commit doing all that stuff might be an improvement for the future though...

PR #1235, opened 2021-07-23
Send MDNS service requests over several messages

### LINKED ISSUE #1234: Zeroconf response incomplete due to response not fitting in MTU
**Describe the bug**
Original issue was discovered in https://github.com/home-assistant/core/issues/52862#issuecomment-885229784

The problem is that now when pyatv asks for a lot of services in the same request, the anticipated response might not fit in the packet MTU (roughly 1450 bytes or so). This makes the sender ignore records, thus making the final response incomplete. In this case everything was present, but only a AAAA record no A record. So I have to split and make several requests, see mentioned issue above for details.

**To Reproduce**
Scan (regular or unicast) for a device producing too large responses. Device is not found.

**Expected behavior**
Device is found

**System Setup (please complete the following information):**
 - OS: any
 - Python: any
 - pyatv: 0.8.0-0.8.1
 - Apple TV: any

**Additional context**
Implicitly introduced in 0.8.0 as many additional services were added.

postlund:
This PR aims to fix #1234 by splitting requests over several messages (instead of just one). So far only unicast scanning has been (partially implemented).#1218 

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1235"></a>

### DIFF
--- a/pyatv/support/mdns.py
+++ b/pyatv/support/mdns.py
@@ -455,11 +455,11 @@ class MulticastDnsSdClientProtocol:  # pylint: disable=too-many-instance-attribu
                 [service.name + "." + service.type for service in services],
                 QueryType.ANY,
             )
-        else:
+        elif query_resp.count >= len(self.queries):
             response = Response(
-                services=services,
+                services=query_resp.parser.parse(),
                 deep_sleep=query_resp.deep_sleep,
-                model=_get_model(services),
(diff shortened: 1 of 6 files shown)


postlund:
Everything implemented and verified in https://github.com/home-assistant/core/issues/52862

Issue #1209, opened 2021-07-13
Add support for a connect strategy


postlund:
**Short feature/function description**
Currently only DMAP and MRP tries to connect to the remote device when calling `pyatv.connect`. The other protocols do that on-demand, when needed. Only DMAP and MRP support persistent connections so far, which is the reason it works like this. One problem though is that a service can fail because of fatal error, like missing/missing credentials or improper configuration (e.g. wrong port). Ideally, this should be caught during initial connect so that problems can be signaled immediately.

My idea is to introduce a "strategy", so a developer can choose what will happen if connecting to a service fails. I initially intend to support two strategies:

1. Fail if any protocols fails to connect
2. Ignore protocols that fail to connect

An API needs to be added to expose which protocol that are actually connected as well as potential exception/error message in case a protocol failed to connect.

Current strategy matches 1 and I would like to keep it like that.

**What needs to be done?**

* Add parameter to `pyatv.connect` accepting a strategy (needs a new constant in `const.py`)
* Connect call in `facade.py` is currently sequential and should be made parallel (call connect on each protocol and collect with gather)
* Add API to `interface.AppleTV` with service status, I.e. which protocols are connected (need to think about this a bit more)

**Is this a breaking change?**
Yes and no. No change from an API perspective, but connect will fail in some cases where it would not before.

**Anything else worth knowing?**
No

Issue #886, opened 2020-12-04
Device State becomes unresponsive after video played via media_player.play_media service until restart Apple TV


1Humza:
**Describe the bug**
Media player device state becomes only reactive to video played via media_player.play_media service.
Ex: Watching Netflix would show, then after playing a video with media_player service, the state only changes to videos played with the media_player service.
Have to restart Apple TV to have the state behave as normal.

**To Reproduce**
Play video with url using media_player.play_media service.

**Expected behavior**
After video played via media_player service, device state should change based on what the Apple TV is doing afterwards. Ex: playing another movie

**System Setup (please complete the following information):**
 - OS: Linux
 - Python: 3.8.5
 - pyatv: 0.7.4
- Apple TV: Apple TV 4K tvOS 14.2

**Additional context**
Maybe HA is only reading from the new AirPlayStreamAPI instance? Does the session fail to close?

postlund:
No metadata is read from AirPlay yet (support for that is a later matter). I'm not sure why this would happen. But are you saying it's the same behavior on iOS, or am I misunderstanding what you wrote?

1Humza:
Yes, it's the same behavior on iOS. My phone shows nothing playing on the Apple TV as well. Both HA and iOS only report that something is playing when I play something using the media_player service after I use it once. Prior to using media_player service, everything shows as normal.

1Humza:
`        try:
            player = await self._player(session_manager.session)
            position = int(kwargs.get("position", 0))
            self._play_task = asyncio.ensure_future(player.play_url(url, position))
            return await self._play_task
        finally:
            self._play_task = None
            await session_manager.close()
            if server:
                await server.close()`
Would the return await self._play_task in this situation cause the function to end before finally:... could be triggered? Could this behavior be caused because the _play_task is never set to None?

postlund:
The `finally` clause is always executed no matter what happens, that's guaranteed by the language. To me it sounds more like a bug in the Apple TV. If it's caused by a dangling connection, try playing something with `play_media` and then restart Home Assistant as that will close all connections. If the problem persists, it's an issue with tvOS.

1Humza:
[...]

PR #1263, opened 2021-08-24
Add support for MRP over AirPlay

postlund:
So, this is a *huge* PR adding support for setting up an MRP channel over AirPlay 2. ~~So far this is just work-in-progress. Most of the foundation is in place (and also implemented the way I want it), but the final integration is just PoC to show that it works.~~ Most things should hopefully be in place to start doing some serious testing. Error handling is area where I might lack some handling, so hoping to get some input in that department.

To test this out, start by pairing to get credentials:

    atvremote -s <ip> --protocol airplay pair

Note that you cannot use old AirPlay credentials as new ones are needed for AirPlay 2. So pair again and save the credentials in a file, e.g. `aplay_creds`.

To try it out, just make sure to load the credentials and it should work:

    atvremote -s <ip> --airplay-credentials `cat aplay_creds` --scan-protocols airplay playing
 
 Add `--debug` to get debug logs (will help me troubleshoot in case of problems).

~~Some caveats are in order. It breaks regular MRP and also AirPlay 1 support (`play_url`).
Supported features, e.g. `atvremote ... features` does not work. It's possible to use `push_updates` (based on simple testing),  it there's some sort of heartbeat I'm not doing so the connection is closed after about 30s.~~

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1263"></a>


azogue:
Hi @postlund, thank you SO MUCH for the extraordinary work you're doing here 🍻

I'm trying this PR and indeed **it works** 💪🚀. Mi setup is one **AppleTV4KGen2** (newer ones) with the first public beta (and also with newer beta2, just updated to continue testing)

Things that are **working ok** after pairing with `--airplay-credentials`:

* Device Info commands (except `build_number`, returning empty)
* Remote button commands
* `push_updates` show new msgs when playing starts on some app
* Power commands (and I don't know if that's new, but now I can power-off/on the TV+HomeCinema via CEC 🤩)
* Metadata info (media_type appear as unknown for some apps like youtube, but overall everything seem fine)  

Things not working:

* Apps commands `launch_app` and `app_list` fail with `NotSupportedError` (don't remember if they were working before for ATV 😅)
* AirPlay command `play_url` fails without exceptions after some time (**debug log attached** for `atvremote -s 192.168.1.138 --airplay-credentials `cat aplay_creds` --scan-protocols airplay play_url=https://youtu.be/KtpeENQk8gI --debug > debug_log_play_url_command.log`)

Feel free to ask for more testing or debug logs, and again, thank you for the effort and for this awesome library 🤗

[debug_log_play_url_command.log](https://github.com/postlund/pyatv/files/7072269/debug_log_play_url_command.log)

postlund:
> Hi @postlund, thank you SO MUCH for the extraordinary work you're doing here 🍻

Hi and thank you (also for testing)! 😊

> I'm trying this PR and indeed **it works** 💪🚀. Mi setup is one **AppleTV4KGen2** (newer ones) with the first public beta (and also with newer beta2, just updated to continue testing)

That's great to hear! It at least means I'm on the right track.

> Things that are **working ok** after pairing with `--airplay-credentials`:

Let's look these over one by one:

> * Device Info commands (except `build_number`, returning empty)

I think the build number is extracted from the MRP zeroconf service which isn't available now, so it will be missing. It's part of the initial `DEVICE_INFO_MESSAGE` however, so it's possible to make it work again. It requires dynamic updates of device info though, as that information is only available after connecting and device info currently only depend on what's available via zeroconf. That's an improvement I can add.

> * Remote button commands

👍

> * `push_updates` show new msgs when playing starts on some app

👍

> * Power commands (and I don't know if that's new, but now I can power-off/on the TV+HomeCinema via CEC 🤩)

That's probably new in tvOS as nothing has changed there. Using the Companion protocol would work in that way before though.

> * Metadata info (media_type appear as unknown for some apps like youtube, but overall everything seem fine)

Yeah, media_type seems to be a bit off. Haven't looked into if that is a bug or just reported wrongly by the device.

> Things not working:
> 
> * Apps commands `launch_app` and `app_list` fail with `NotSupportedError` (don't remember if they were working before for ATV 😅)

You need to pair companion and provide credentials for those to work. There has been some problems in the early betas but should work ok latest-and-greatest.

> * AirPlay command `play_url` fails without exceptions after some time (**debug log attached** for `atvremote -s 192.168.1.138 --airplay-credentials `cat aplay_creds` --scan-protocols airplay play_url=https://youtu.be/KtpeENQk8gI --debug > debug_log_play_url_command.log`)

This seems to be a long standing and very annoying bug (as you have to restart the device to recover). But I would say it's a bug I tvOS that apple doesn't seem to fix. More details in https://github.com/postlund/pyatv/issues/886.

> Feel free to ask for more testing or debug logs, and again, thank you for the effort and for this awesome library 🤗
> 
> [debug_log_play_url_command.log](https://github.com/postlund/pyatv/files/7072269/debug_log_play_url_command.log)

I tried leaving `push_updates` running for about 12h yesterday before getting an exception. The connection broke but as I didn't have any debug logs enabled for that, I couldn't troubleshoot. I haven't added support for device updates yet (calling a listener to perform an action when the connection breaks), so it's expected right now. But testing long-term would be appreciated to find stability problems.

postlund:
I have some things that I will push later tonight. Some things that I have fixed/changed:

* Better handling of connection errors, making sure everything is torn down and properly signaled to the device listener.
* In current code, protocols to be set up are first added to a list and then connected to. Resolution is performed in the first step, so that regular MRP and MRP in AirPlay aren't added simultaneously for instance (it doesn't make sense to connect to the same protocol twice). There's a problem with this however. Since AirPlay happens to be added first, MRP will be added by AirPlay before regular MRP (assuming the device has remote control support), meaning that regular MRP is just ignored. If the remote control channel fails to set up, which happens if no AirPlay credentials are provided, then MRP is completely ignored. So just doing `atvremote -s <ip> playing` will not work anymore, as the MRP-over-AirPlay tunnel will fail and regular MRP is ignored. I've changed it so that all protocols are first added and then connected one by one, ignoring a protocol that has already been set up successfully. This provides better backwards compatibility and is also good preparation for supporting #1209.
* RAOP will only read credentials from the RAOP service now (`--raop-credentials`) and pairing has been added, which will do legacy pairing (for now). Using the credentials from AirPlay will in most cases not work as they are for HAP, which requires encryption and I have not implemented support for that yet.
* Other minor bug fixes and some very welcome code clean ups.

postlund:
I have pushed what I believe might be the last changes of this. Hopefully it works quite well now.

postlund:
I will merge this now and continue working with other things that needs to be done. I will soon also open an issue with a general request for testing the upcoming 0.9.0 release to iron out bugs.

Issue #1280, opened 2021-09-03
Add official support for python 3.10


postlund:
**What feature would you like?**
Add python 3.10 to list of supported python versions. Final version is not out yet, but I can run with rc1 for now.

**Describe the solution you'd like**
Should just be added to `setup.py` and `tests` workflow. Hopefully it works with no particular changes, but I need to check if something has been deprecated. I know that the `loop` argument has been removed from asyncio functions, so I need to check I don't have any of those left.

**Any other information to share?**
No

PR #1281, opened 2021-09-03
gha: Run tests on python 3.10

postlund:
Relates to #1280


postlund:
Seems to be general tool chain instability and really long install times (likely because of missing wheels and prebuilt packages), so I'm putting this off for a while until things improve.

PR #1289, opened 2021-09-07
gha: Create GitHub release when tag is pushed

postlund:
Today I manually create releases when I make a release, something that
should be automatic. This is an attempt to create automatic releases
when a tag is pushed. It is assumed a tag is always to be treated as a
release. As a test, releases will be marked as draft until I know that
it works.

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1289"></a>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,58 @@
+name: Create GitHub release
+
+on:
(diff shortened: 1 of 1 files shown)


postlund:
Making it another way...

PR #1316, opened 2021-09-17
Refactor conf module and add requires_password and pairing_requirement

postlund:
This is WIP, so not done yet... I've refactored how services are structured altogether and now officially deprecated all service specific implementations (e.g. `MrpService`). Anyone that needs to create a config manually should use `conf.ManualService` instead.

There's now a function in core that lets protocol implementations add information to services during scanning, e.g. if a password or pairing is needed.

So far only password detection has been implemented for AirPlay and RAOP. The remaining parts is next in line. Scanning will tell if a password is required:

```raw
      Name: Office
   Model/SW: HomePodMini tvOS 14.7
    Address: 10.0.10.84
        MAC: xxx
 Deep Sleep: True
Identifiers:
 - xxx
 - xxx
Services:
 - Protocol: Companion, Port: 49152, Credentials: None, Requires Password: False, Password: None
 - Protocol: AirPlay, Port: 7000, Credentials: None, Requires Password: True, Password: None
 - Protocol: RAOP, Port: 7000, Credentials: None, Requires Password: True, Password: None
```

Output format for services is getting rather verbose now. I think I have to re-think that.

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1316"></a>


postlund:
This turned out to be a much bigger change than I anticipated (like most things nowadays). But it feels good that it's done now. It will hopefully make it easier adding new kinds of metadata in the future.

Issue #1372, opened 2021-10-05
Scan works, nothing else does


johnlcoxjr:
### Describe the bug

I was using this before tvOS 15 and it worked fine.  After the update, my implementation stopped working.  Scan works and can find all of the AppleTV's.  However the one running tvOS 15 doesn't respond to any command (other than scan), and always returns "Could not find any Apple TV on current network".

atvremote --scan-hosts 192.168.50.20,192.168.50.21,192.168.50.22,192.168.50.23 scan
Scan Results
========================================
       Name: Family Room 2
   Model/SW: Apple TV 4K (gen2), tvOS 15.0
    Address: 192.168.50.20
        MAC: 58:F3:49:D8:2A:70
 Deep Sleep: False
Identifiers:
 - 58:F3:49:D8:2A:70
 - 58F349D82A70
Services:
 - Protocol: Companion, Port: 49153, Credentials: None, Requires Password: False, Password: None, Pairing: Mandatory
 - Protocol: AirPlay, Port: 7000, Credentials: None, Requires Password: False, Password: None, Pairing: Mandatory
 - Protocol: RAOP, Port: 7000, Credentials: None, Requires Password: False, Password: None, Pairing: Mandatory

### Error log

atvremote -i 58F349D82A70 --protocol companion pair
2021-10-05 11:18:19 ERROR [pyatv.core.mdns]: failed to add listener for 192.168.50.100
Traceback (most recent call last):
  File "p:\platformio\apple-tv\venv\lib\site-packages\pyatv\core\mdns.py", line 522, in multicast
    await protocol.add_socket(net.mcast_socket(str(addr)))
  File "p:\platformio\apple-tv\venv\lib\site-packages\pyatv\support\net.py", line 41, in mcast_socket
    sock.setsockopt(
OSError: [WinError 10022] An invalid argument was supplied
2021-10-05 11:18:19 ERROR [pyatv.core.mdns]: failed to add listener for 127.0.0.1
Traceback (most recent call last):
  File "p:\platformio\apple-tv\venv\lib\site-packages\pyatv\core\mdns.py", line 522, in multicast
    await protocol.add_socket(net.mcast_socket(str(addr)))
  File "p:\platformio\apple-tv\venv\lib\site-packages\pyatv\support\net.py", line 41, in mcast_socket
    sock.setsockopt(
OSError: [WinError 10022] An invalid argument was supplied
[...]

PR #1376, opened 2021-10-06
net: Set multicast options after bind

postlund:
Relates to #1372

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1376"></a>


postlund:
Did not work.

PR #1583, opened 2022-01-05
Add support for scanning with AsyncZeroconf

### LINKED ISSUE #1570: Version 2.2.0 of hass-atv-beta fails to communicate with Apple TV 4K
### Describe the bug

Upon upgrading from `v1.8.1` of hass-atv-beta to `v2.2.0`, I'm unable to communicate with my Apple TV. Downgrading to `v2.1.0` solves the problem. The Apple TV is on a separate VLAN, but mDNS relay is enabled, and this has never been a problem before. Additionally, `v2.1.0` works fine!

### Error log

n/a

### How to reproduce the bug?

1. Install `v2.2.0` of hass-atv-beta
1. Successfully complete pairing
1. Note in the Home Assistant UI that OS version, model number, etc isn't populated.
1. Also note that in the Home Assistant UI, no play status information is conveyed, and turn on/turn off no longer works

### What is expected behavior?

After pairing, communication with the Apple TV should proceed as normal.

### Operating System

Linux (Debian)

### Python

3.9

### pyatv

Unknown (whatever hass-atv-beta v2.2.0 uses)

### Device

Apple TV 4K First Gen tvOS 15.1.1

### Additional context

Downgrading to v2.1.0 solves the problem.

bdraco:
Fixes #1570

Home Assistant changes: https://github.com/bdraco/home-assistant/tree/zc_scanner_atv

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1583"></a>


pyatv/__init__.py
```diff
@@ -41,10 +52,13 @@ def _should_include(atv):
         return True
 
     scanner: BaseScanner
-    if hosts:
-        scanner = UnicastMdnsScanner([IPv4Address(host) for host in hosts], loop)
+    if async_zeroconf:
```
postlund, line 55: Do you see any benefits of keeping any of the old implementations? I would prefer if we could just remove them altogether and just use your new one instead. It would remove so much code and so many obnoxious tests...

bdraco: A service browser has to be running for the types for the zeroconf implementation to work. 

Its easy enough to start one and then kill it off at the end of the scan, but I didn't want to bite off too much in a single PR.

bdraco: Also I didn't implement knocking in the zc scanner.  Everything seemed to work fine without it.  If its needed, I'll do another turn with it later.

postlund: Yeah, I expect that we create one internally and close it when we are done. It's the best compromise I would say.

Port knocking is only relevant for devices that goes to deep sleep. I have only encountered that with my Apple TV 3, where the port knocking feature is needed.

postlund: And also, since the changes you are making does not alter any public interface we can let mine and your implementation live side by side and then we just remove my stuff after merging your changes. Just to keep it clean-ish.

bdraco: > Port knocking is only relevant for devices that goes to deep sleep. I have only encountered that with my Apple TV 3, where the port knocking feature is needed.

Do you still have one of these for testing? I know they are EOL but it would be nice to test with it if we can.

bdraco:
Manual retest looks good. 

Do you see anything that stands out before I start writing tests for this?

tests/core/test_scan.py
```diff
@@ -34,3 +48,191 @@ def test_unique_identifiers(unique_id_mock, response):
     assert "id2" == next(identifiers)
     unique_id_mock.assert_called_with("_service2._tcp.local", "service2", {"c": "d"})
     assert not next(identifiers, None)
+
+
+@pytest.mark.asyncio
+async def test_scan_with_zeroconf():
+    if not hasattr(asyncio, "get_running_loop"):
+        # until python 3.6 is removed, its EOL anyways
+        asyncio.get_running_loop = asyncio.get_event_loop
```
bdraco, line 57: This needs to get removed, but I didn't want to assume its ok to drop python3.6 support

zeroconf no longer supports 3.6 since it doesn't support EOL py3 versions

postlund:
I will try to do a code review later tonight.

tests/core/test_scan.py
postlund: I will deprecate python 3.6 according to #1587. Hopefully I can do that today.

pyatv/__init__.py
```diff
@@ -27,8 +33,13 @@ async def scan(
     identifier: Optional[Union[str, Set[str]]] = None,
     protocol: Optional[Union[Protocol, Set[Protocol]]] = None,
     hosts: List[str] = None,
+    async_zeroconf: Optional[AsyncZeroconf] = None,
```
postlund, line 36:
```suggestion
    zeroconf: Optional[AsyncZeroconf] = None,
```

I like it short (type hint tells us it's async).

bdraco: Does aiozc work here?

since there is a zeroconf object inside the async version it makes it confusing to name it the same

pyatv/core/scan.py
```diff
@@ -288,3 +311,171 @@ def _end_if_identifier_found(self, response: mdns.Response):
...
+        aiozc: AsyncZeroconf,
+        hosts: Optional[List[IPv4Address]] = None,
+    ) -> None:
+        """Initialize a new scanner."""
+        super().__init__()
+        self.aiozc = aiozc
+        self.zeroconf = aiozc.zeroconf
+        self.hosts: Set[str] = set(str(host) for host in hosts) if hosts else set()
+
+    async def _async_services_by_addresses(
```
postlund, line 370: Minor thing, but as pyatv is "fully async" there's really no need for the `_async` prefix to spell out the context. I don't care, but just to mention it (I don't use it anywhere else).

pyatv/core/scan.py
```diff
@@ -288,3 +311,171 @@ def _end_if_identifier_found(self, response: mdns.Response):
...
+                    ),
+                    model=model,
+                )
+            )
+
+    async def process(self, timeout: int) -> None:
+        """Start to process devices and services."""
+        zc_timeout = timeout * 1000
+        services_by_address = await self._async_services_by_addresses(zc_timeout)
+        atv_services_by_address: Dict[str, List[mdns.Service]] = {}
```
postlund, line 448:
```suggestion
        dev_services_by_address: Dict[str, List[mdns.Service]] = {}
```
Or something similar. As pyatv supports more devices than Apple TVs nowadays, we should refrain from using terms like "atv". There are remaining occurrences and I try to remove them along the way.

pyatv/core/scan.py
```diff
@@ -288,3 +311,171 @@ def _end_if_identifier_found(self, response: mdns.Response):
...
+        name_by_address: Dict[str, str] = {}
+        for address, services in services_by_address.items():
+            if self.hosts and address not in self.hosts:
+                continue
+            atv_services = []
+            for service in services:
+                atv_type = service.type[:-1]
```
postlund, line 455:
```suggestion
                service_type = service.type[:-1]
```

pyatv/core/scan.py
```diff
@@ -288,3 +311,171 @@ def _end_if_identifier_found(self, response: mdns.Response):
...
+        name_by_address: Dict[str, str],
+    ):
+        """Process and callback each aggregated response to the base handler."""
+        for address, atv_services in atv_services_by_address.items():
+            model = None
+            name_for_address = name_by_address.get(address)
+            if name_for_address is not None:
+                possible_model = name_to_model.get(name_for_address)
+                if possible_model:
+                    model = possible_model
```
postlund, line 432: Why not just:
```python
model = name_to_model.get(name_for_address)
```
?

tests/core/test_scan.py
```diff
@@ -1,21 +1,238 @@
 """Unit tests for scan module."""
```
postlund, line 1: I have no looked at these tests for now.

postlund (changes requested):

pyatv/__init__.py
postlund: Yeah, that is fine!

bdraco: ea6f728

pyatv/helpers.py
```diff
@@ -70,10 +70,25 @@ def get_unique_id(
     if service_type == AIRPLAY_SERVICE:
         return properties.get("deviceid")
     if service_type == RAOP_SERVICE:
-        return service_name.split("@", maxsplit=1)[0]
+        return raop_unique_id_from_service_name(service_name)
     return None
 
 
+def raop_unique_id_from_service_name(service_name: str) -> str:
```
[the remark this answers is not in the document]
bdraco: 74f52af

pyatv/core/scan.py
```diff
@@ -42,7 +54,18 @@
 ]
 
 DEVICE_INFO: str = "_device-info._tcp.local"
+DEVICE_INFO_TYPE: str = f"{DEVICE_INFO}."
 SLEEP_PROXY: str = "_sleep-proxy._udp.local"
+SLEEP_PROXY_TYPE: str = f"{SLEEP_PROXY}."
+RAOP_TYPE: str = f"{RAOP_SERVICE}."
```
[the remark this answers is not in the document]
bdraco: These are only used to build `NAME_USED_FOR_DEVICE_INFO`

`_device_info` is a special case as it only publishes TXT records so there is no PTR to browse

pyatv/core/scan.py
```diff
@@ -288,3 +311,171 @@ def _end_if_identifier_found(self, response: mdns.Response):
         return self.identifier and not self.identifier.isdisjoint(
             set(get_unique_identifiers(response))
         )
+
+
+def _extract_service_name(info: AsyncServiceInfo) -> str:
+    return info.name[: -(len(info.type) + 1)]
+
+
+def _device_info_name(info: AsyncServiceInfo) -> Optional[str]:
```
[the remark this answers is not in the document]
bdraco: I'm looking for the name to map it for the `device_info` here instead of the unique id

pyatv/core/scan.py
```diff
@@ -288,3 +311,171 @@ def _end_if_identifier_found(self, response: mdns.Response):
...
+        return None
+    short_name = _extract_service_name(info)
+    if info.type == RAOP_TYPE:
+        return raop_name_from_service_name(short_name)
+    if info.type == SLEEP_PROXY_TYPE:
+        return sleep_proxy_name_from_service_name(short_name)
+    return short_name
+
+
+def _first_non_link_local_or_v6_address(addresses: List[bytes]) -> Optional[str]:
```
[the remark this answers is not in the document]
bdraco: I think this is bad naming on my part as its excluding ipv6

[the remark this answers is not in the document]
bdraco: 0ddfde2358ef96b446690c8c42352534bfd26470

pyatv/core/scan.py
bdraco: 96827d4fa480cdd707a8bae09dc4a54efc8abb6a

pyatv/core/scan.py
bdraco: 44b128fa5ee0418ec8ed40e15f5e4a7d23c9a105

pyatv/core/scan.py
bdraco: 7082eb6037d3afcfe18c403a5f507303ec3f89fc

pyatv/core/scan.py
bdraco: It mattered in a previous iteration before I broke the function apart but doesn't now

2bc9acc5b1d1c2b31eeb3529ebecd05e99ecc3c8

pyatv/core/scan.py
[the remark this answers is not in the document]
bdraco: Figured out a way to do it without making it too much bigger

51ea291ac4d23a439f1b940a88513360ceb49f38

pyatv/core/scan.py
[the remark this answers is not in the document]
bdraco: This has been removed

pyatv/core/scan.py
```diff
@@ -288,3 +312,162 @@ def _end_if_identifier_found(self, response: mdns.Response):
...
+
+def _extract_service_name(info: AsyncServiceInfo) -> str:
+    return info.name[: -(len(info.type) + 1)]
+
+
+def _first_non_link_local_or_non_v6_address(addresses: List[bytes]) -> Optional[str]:
+    """Return the first ipv6 or non-link local ipv4 address."""
+    for address in addresses:
+        ip_addr = ip_address(address)
+        if not ip_addr.is_link_local or ip_addr.version == 6:
```
bdraco, line 325:
```suggestion
        if not (ip_addr.is_link_local or ip_addr.version == 6):
```

pyatv/core/scan.py
```diff
@@ -288,3 +312,162 @@ def _end_if_identifier_found(self, response: mdns.Response):
...
             set(get_unique_identifiers(response))
         )
+
+
+def _extract_service_name(info: AsyncServiceInfo) -> str:
+    return info.name[: -(len(info.type) + 1)]
+
+
+def _first_non_link_local_or_non_v6_address(addresses: List[bytes]) -> Optional[str]:
+    """Return the first ipv6 or non-link local ipv4 address."""
```
bdraco, line 322:
```suggestion
    """Return the first non ipv6 or non-link local ipv4 address."""
```

bdraco:
Updated the HASS branch to use the `aiozc` name as well

postlund:
I feel that I'm still not satisfied with the `_device-info` handling 🙁 It leaks irrelevant and quite internal details into the protocol logic. Since that service only contains the hardware device model, can't we just fetch all of them for a device based on IP address and just pick one? It should contain the same value for all services from a particular device anyway.

bdraco:
> I feel that I'm still not satisfied with the `_device-info` handling 🙁 It leaks irrelevant and quite internal details into the protocol logic. Since that service only contains the hardware device model, can't we just fetch all of them for a device based on IP address and just pick one? It should contain the same value for all services from a particular device anyway.

How to you propose to construct the query for the device info since there is no PTR published for it?

bdraco:
There are also no A records for the _device-info type so we can't link it by address

bdraco:
I'll have to think about this one.  I don't have a good solution for https://github.com/postlund/pyatv/pull/1583#issuecomment-1007647181

bdraco:
I can be made more efficient by deriving device-info name in the `_services_by_addresses`.  Unfortunately thats a major refactor so it will take a few hours to implement which I don't have right now

bdraco:
Since `device-info` records don't have a PTR record, they have to be handled differently.  

I may be reading the existing code wrong, but AFAICT the current implementation is relying on the device responding with `device-info` records in the additional section.

While they SHOULD be there, they aren't required as additionals are an optimization and since there is no query that is explicitly asks for `DEVICENAME._device-info._tcp.local.` it relies on mDNSResponder not running out of space in the responses packet.  [RFC6762](https://lxr.openwrt.org/source/mdnsd/rfc6762.txt#L1033) 

From `mDNSResponder-1310.80.1` `mDNSCore/mDNS.c`

```c
        // Add additionals, if there's space
        while (ResponseRecords && !ResponseRecords->NR_AnswerTo)
        {
            rr = ResponseRecords;
            if (rr->resrec.RecordType & kDNSRecordTypeUniqueMask)
                rr->resrec.rrclass |= kDNSClass_UniqueRRSet;        // Temporarily set the cache flush bit so PutResourceRecord will set it
            newptr = PutResourceRecord(&m->omsg, responseptr, &m->omsg.h.numAdditionals, &rr->resrec);
            rr->resrec.rrclass &= ~kDNSClass_UniqueRRSet;           // Make sure to clear cache flush bit back to normal state

            if (newptr) responseptr = newptr;
            if (newptr && m->omsg.h.numAnswers) rr->RequireGoodbye = mDNStrue;
            else if (rr->resrec.RecordType & kDNSRecordTypeUniqueMask) rr->ImmedAnswer = mDNSInterfaceMark;
            ResponseRecords = rr->NextResponse;
            rr->NextResponse    = mDNSNULL;
            rr->NR_AnswerTo     = mDNSNULL;
            rr->NR_AdditionalTo = mDNSNULL;
        }

```

bdraco:
I've been running this in production for a few days, and everything seems fine.

Eightplant:
> I've been running this in production for a few days, and everything seems fine.

If you wanted to instruct me how to install the modified code I can give it a go.

bdraco:
https://developers.home-assistant.io/docs/development_environment?_highlight=venv#setup-local-repository

In your Home Assistant environment :

- Checkout this PR with `gh pr checkout 1583` using GH cli https://cli.github.com/
- Install it with `pip3 install --upgrade .`
- Checkout this branch https://github.com/bdraco/home-assistant/tree/zc_scanner_atv 
- Start hass

postlund:
> Since `device-info` records don't have a PTR record, they have to be handled differently.
> 
> I may be reading the existing code wrong, but AFAICT the current implementation is relying on the device responding with `device-info` records in the additional section by reducing number of types in the query to 3 to hope that the answers all fit in the response. Its also allowed for unrelated aggregated responses to be in the response since https://lxr.openwrt.org/source/mdnsd/rfc6762.txt#L599 so it seems like its possible for unexpected data to also be returned in the response
> 
> While they SHOULD be there, they aren't required as additionals are an optimization and since there is no query that is explicitly asking for `DEVICENAME._device-info._tcp.local.`, as it relies on mDNSResponder not running out of space in the response packet and not getting unlucky that other responses get aggregated. [RFC6762](https://lxr.openwrt.org/source/mdnsd/rfc6762.txt#L1033)
> 
> From `mDNSResponder-1310.80.1` `mDNSCore/mDNS.c`
> 
> ```c
>         // Add additionals, if there's space
>         while (ResponseRecords && !ResponseRecords->NR_AnswerTo)
>         {
>             rr = ResponseRecords;
>             if (rr->resrec.RecordType & kDNSRecordTypeUniqueMask)
>                 rr->resrec.rrclass |= kDNSClass_UniqueRRSet;        // Temporarily set the cache flush bit so PutResourceRecord will set it
>             newptr = PutResourceRecord(&m->omsg, responseptr, &m->omsg.h.numAdditionals, &rr->resrec);
>             rr->resrec.rrclass &= ~kDNSClass_UniqueRRSet;           // Make sure to clear cache flush bit back to normal state
> 
>             if (newptr) responseptr = newptr;
>             if (newptr && m->omsg.h.numAnswers) rr->RequireGoodbye = mDNStrue;
>             else if (rr->resrec.RecordType & kDNSRecordTypeUniqueMask) rr->ImmedAnswer = mDNSInterfaceMark;
>             ResponseRecords = rr->NextResponse;
>             rr->NextResponse    = mDNSNULL;
>             rr->NR_AnswerTo     = mDNSNULL;
>             rr->NR_AdditionalTo = mDNSNULL;
>         }
> ```

Yes, I'm looking for it in the response and save it in case it's there (I think Apple normally include this). It's just pure best-effort and if it's not found, then it's not the end of the world. Device model can be derived from other properties in the various services, this is just one way. If the code the be simplified so we can remove the protocol logic that is added now, that would be preferred IMHO. If it's hard/not feasible, then I guess I will have to accept this solution as I wand to merge this as soon as I can.

I'm a bit behind on pyatv right now, but I'm gonna start working through the backlog the next few days and make a release soon, so we can this and a few bug fixes in before next release of Home Assistant.

tests/core/test_scan.py
postlund: #1587 is merged now.

postlund:
> > If the code the be simplified so we can remove the protocol logic that is added now, that would be preferred IMHO. If it's hard/not feasible, then I guess I will have to accept this solution as I wand to merge this as soon as I can.
> 
> The current design is to figure out the name to ask for so we can include all the queries in one-shot which makes is really efficient and reliable since zeroconf knows how to sort out which ones are missing and ask the questions. We should _always_ be able to get the device info with the current state of this PR (unless the device is not responding)
> 
> I can remove the support for finding the `device-info` if that preferred.

Ok! Leave it as it is, I'll make a final review tomorrow morning and after that we can merge this beast.

pyatv/protocols/dmap/__init__.py
```diff
@@ -612,12 +612,12 @@ def hscp_service_handler(
...
 
 
-def scan() -> Mapping[str, ScanHandler]:
+def scan() -> Mapping[str, ScanHandlerDeviceInfoName]:
     """Return handlers used for scanning."""
     return {
-        "_appletv-v2._tcp.local": homesharing_service_handler,
-        "_touch-able._tcp.local": dmap_service_handler,
-        "_hscp._tcp.local": hscp_service_handler,
+        "_appletv-v2._tcp.local": (homesharing_service_handler, lambda _: None),
```
postlund, line 618: I guess these services should do some kind of extraction as well and not just return `None`?

bdraco: These devices don't have enough information in the name to derive the device info name so `None` is expected

postlund: Ok, great! Then we are all set for now.

postlund:
Ok, so what remains now is to use this implementation with regular multicast and unicast scan as well. I guess most changes concerns setting up (and tearing down) a ServiceBrowser and then reuse the code from this PR.

bdraco:
I'll do another turn on this to replace the multicast/unicast scans once we get a full cycle of it in Home Assistant in case there are any corner cases to address. 

It will be a bit more complex since we won't have a service browser running but I can reuse most of whats there now.

postlund:
I expect that it would be best to make a new release of pyatv to get this out before making those changes, no? I would be OK with that, just need to settle a few more other outstanding bugs first.

bdraco:
> I expect that it would be best to make a new release of pyatv to get this out before making those changes, no?

Yes. It would be good to do a release, bump in https://github.com/home-assistant/core/pull/64528, and merge, so we can get feedback for next week's Home Assistant beta.

>  I would be OK with that, just need to settle a few more other outstanding bugs first.

👍

Issue #649, opened 2020-05-10
Commands take ~8s to respond


parautenbach:
I'm testing out the latest version in anticipation of an updated HASS integration. Thank you so much for the work on this. 

While experimenting with the commands, I see every command I've tried takes about 8s to respond. I'm not sure whether it's supposed to be like this or whether it's still beta code. If this isn't supposed to be the case, I'd be happy to log a bug and submit info that would help to debug this.

postlund:
That's not how it's supposed to be. Some logs would likely help me troubleshooting. There are some instructions in the hass-atv-beta repo readme file (at the bottom).

parautenbach:
Cool, I'll check the docs and hopefully submit you a log a bit later.

I did spot the section on logging within HASS. I should perhaps have mentioned that I'm using the `atvremote` command line tool.

postlund:
Ah, if you are using atvremote then it's another matter. What hardware are you running on?

parautenbach:
Right, I was wondering whether it might make a difference (guessing it establishes a new connection with a lot of handshaking overhead every time). 

I found the [logging docs](https://github.com/postlund/pyatv/blob/master/docs/documentation/atvremote.md) for `atvremote`.

postlund:
Every time you run a command a new connection (with encryption and all that) is set up. That part is relatively fast though, even on low-end hardware. Loading all libraries into memory from a slow SD-card adds some penalty when it comes to loading times. Biggest culprit is however scanning to find the device, which is done for every command and set to 3s. If you are on the same subnet as the device and have assigned it a static IP, you can utilize unicast scanning by providing `-s x.x.x.x`. It will be a lot faster. You can also omit `--id` when doing so if you like.

parautenbach:
Ah, thanks for that. In this case, I'm running it on a Pi 3B+ with an SD. With your suggestions, it knows a few seconds off the call. 

```
(venv) xxxxx:~/pyatv $ atvremote --debug --verbose -s www.xxx.yyy.zzz --mrp-credentials `cat credentials.txt` power_state
2020-05-11 12:13:36 DEBUG: Knocking at port 3689 on www.xxx.yyy.zzz
2020-05-11 12:13:36 DEBUG: Knocking at port 7000 on www.xxx.yyy.zzz
2020-05-11 12:13:36 DEBUG: Knocking at port 49152 on www.xxx.yyy.zzz
[...]

PR #1645, opened 2022-02-04
simple example json api

twocolors:
Hello, this is a simple example of a JSON API server, it will be useful if you have services in Docker and want to communicate between services (Docker)

As an example, when a Node-Red is in the docker and you need to work with pyatv

Added here (send pull request), maybe someone will help to make it better or correct errors

Thank you

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1645"></a>

### DIFF
--- a/examples/json-api.py
+++ b/examples/json-api.py
@@ -73,41 +73,39 @@ class PushPrinter(DeviceListener, PushListener, PowerListener):
     def connection_lost(self, exception: Exception) -> None:
         """Call when connection was lost."""
         self._remove()
+        payload = output(False, exception=exception, values={"connection": "lost"})
+        self._send_json(payload)
 
     def connection_closed(self) -> None:
         """Call when connection was closed."""
         self._remove()
+        payload = output(True, values={"connection": "closed"})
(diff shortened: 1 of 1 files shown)


postlund:
Sorry about the radio silence! I will have a look at it as soon as I get a few minutes and get back to you.

twocolors:
ping

NorthernMan54:
@twocolors Do you more details on the JSON API or a sample client ?  I was trying to see if it can solve this issue https://github.com/postlund/pyatv/issues/649 and our use case with Homebridge where it is very slow ?

twocolors:
@NorthernMan54 while there is no more information, I'm waiting for @postland to look at the code and say a comment.

But the main idea is to manage ATV by GET/WS

Perhaps this will help in solving your problem because there will always be a connection

NorthernMan54:
Not being a python person, I did get it running, and was able to scan my devices, but couldn’t get connect to work, as i couldn’t figure out the credentials.  I had figured you must have something I could use to see if it works.

PR #1654, opened 2022-02-10
A few more fields for DeviceInfoMessage

systemcrash:
Not sure if you're down with "unknown" in the name, but here ya go. 

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/1654"></a>


systemcrash:
Those extra fields are sent by my iOS 15.3 device.

postlund:
Not a big fan of unknowns (I usually just add a comment about it). But I have looked up the correct names and you can update to them if you want:

* 40 = supportsMultiplayer
* 41 = routingContextID
* 42 = airPlayGroupID
* 43 = systemBooksApplication

postlund:
You also need to run `./scripts/protobuf.py generate` to re-generate the code.

systemcrash:
My script run seems to want to update every single protobuf file, but the changes seemed inconsequential. Excluded for now.

postlund:
I might have bumped protobuf version before without re-running that command, so it is probably expected.

postlund:
@systemcrash If you could re-generate the protobuf messages and push an update, I'll merge this.

postlund:
I opened #1778 so that I can merge these changes.

PR #1756, opened 2022-06-19
Switch to using ChaCha20Poly1305Reusable for encryption

bdraco:
I built this for HomeKit but I realized its used here heavily as well
when profiling since I was seeing the ctx get created over and over
even after the HomeKit changes.  It also seems to fix the Home app
stalling out on iOS 16 betas when using them with Home Assistant's
HomeKit integration.

The ChaCha20Poly1305 that comes with cryptography recreates the
ctx every time encrypt is called. Since we call encrypt in
a loop, we can avoid this overhead by using ChaCha20Poly1305Reusable
instead as we are doing everything in the same thread.

Since `pyatv` is sending back and forth encrypted bytes frequently
with the `heartbeater` this reduces the memory churn and cpu
load if you have a few apple tvs or homepods

### DIFF
--- a/base_versions.txt
+++ b/base_versions.txt
@@ -1,6 +1,7 @@
 aiohttp==3.1.0
 bitarray==2.1.2
-cryptography==2.6
+cryptography==37.0.2
+chacha20poly1305-reuseable==0.0.3
 ifaddr==0.1.7
 mediafile==0.8.1
 miniaudio==1.45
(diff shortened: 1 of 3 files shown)


bdraco:
@postlund Any chance you can take a look at this before Home Assistant beta?

postlund:
@bdraco I'll try to take a look at it by tonight.

bdraco:
Thanks 👍

PR #1810, opened 2022-07-28
Switch out asyncio.wait_for with async_timeout in hot paths

bdraco:
I keep forgetting to do this, and I keep using `asyncio.wait_for` in places where
`async_timeout` would be better, but Martin reminded me that we should be
using `async_timeout` where we can in https://github.com/home-assistant/core/pull/75848
so it made me remember here as well.

 `async_timeout` is already a dep of aiohttp which uses `async_timeout`
  since its much faster than `asyncio.wait_for`

https://pypi.org/project/async-timeout/
Screen Shot 2022-07-28 at 7 29 53 AM

### DIFF
--- a/base_versions.txt
+++ b/base_versions.txt
@@ -1,4 +1,5 @@
 aiohttp==3.8.1,<5
+async_timeout==4.0.2
 bitarray==2.1.2
 cryptography==36.0.2
 chacha20poly1305-reuseable==0.0.3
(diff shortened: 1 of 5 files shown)


bdraco:
With the change, the encryption overhead is now the top on the list instead of the asyncio.wait overhead which is what we would expect
Screen Shot 2022-07-28 at 7 39 44 AM

postlund:
Looks good to me 👍 I will try to fix the `set_volume` bug and make a release ASAP.

bdraco:
> Looks good to me 👍 I will try to fix the `set_volume` bug and make a release ASAP.

Thanks

postlund:
@bdraco 0.10.3 will be ready soon, I will however not have the time right now to bump it. Feel free to do so if you have time (it fixes a few open issues as well).

PR #1829, opened 2022-08-20
Add support for URL param to LaunchApp

chapados:
This change adds support for passing the optional, undocumented `_urlS`
parameter to `_launchApp`. Passing this parameter enables deep linking
into at least some apps installed on the AppleTV. I discovered this
while searching for a way to launch content directly from a deep link
into the AppleTV+ app.

The first clue that this param exists was that passing a valid URL in
 place of a bundleID returns the error message "No valid bundleID or
URL to LaunchApp". Searching for strings in the rootfs of the tvos 15.6
image for AppleTV5,3 (build 19M65) suggested that the `rapportd` daemon
is responsible for handling companion protocol requests and returning
this error message.

A hidden `_urlS` string key is present in the `rapportd` daemon located near
`_bundleID`:

```
$ strings rapportd | grep -C 5 bundleID
_launchApp
v32@?0@"NSDictionary"8@"NSDictionary"16@?<v@?@"NSDictionary"@"NSDictionary"@"NSError">24
_sessionStart
_sessionStop
v40@?0@"RPConnection"8@"NSDictionary"16@"NSDictionary"24@?<v@?@"NSDictionary"@"NSDictionary"@"NSError">32
_bundleID
RPLaunchApp
_urlS
RPOpenURL
-[RPCompanionLinkDaemon _miscHandleLaunchAppRequest:responseHandler:]_block_invoke
```

Disassembly of the object code reveals the following logic in the
`-[RPCompanionLinkDaemon _miscHandleLaunchAppRequest:responseHandler:]` method:

```
(pseudocode)
if (_bundleID):
    launch_app(_bundleID)
elif (_urlS):
    open_url(_urlS)
else:
    error("No valid bundleID or URL to LaunchApp")
```

If a bundleID is passed to `_launchApp`, it takes precendce. The implementation
provided here does not attempt to enforce this logic on the client. Instead,
params are passed directly through to the server so that there is only one
source of truth for the behavior. The current server behavior is represented
in the updated companion protocol tests.

### DIFF
--- a/docs/api/pyatv/interface.html
+++ b/docs/api/pyatv/interface.html
@@ -426,7 +426,7 @@ all its features.</p>
 </dd>
 <dt id="pyatv.interface.Apps.launch_app">
 <code class="name flex">
-<span>async def <span class="ident">launch_app</span></span>(<span>self, bundle_id: str) -> None</span>
+<span>async def <span class="ident">launch_app</span></span>(<span>self, bundle_id: Optional[str], url: Optional[str]) -> None</span>
 </code>
 </dt>
 <dd>
@@ -434,7 +434,7 @@ all its features.</p>
 <span>Feature: <a title="pyatv.const.FeatureName.LaunchApp" href="const#pyatv.const.FeatureName.LaunchApp">FeatureName.LaunchApp</a>,</span>
 <span>Supported by: <a title="pyatv.const.Protocol.Companion" href="const#pyatv.const.Protocol.Companion">Protocol.Companion</a></span>
 </div>
-<section class="desc"><p>Launch an app based on bundle ID.</p></section>
+<section class="desc"><p>Launch an app based on bundle ID or URL.</p></section>
 <div class="git-link-div"><a href="https://github.com/postlund/pyatv/blob/master/pyatv/interface.py#L679-L682" class="git-link">Browse git</a></div>
 </dd>
 </dl>
--- a/docs/development/apps.md
+++ b/docs/development/apps.md
@@ -32,3 +32,9 @@ To launch an app, use its bundle identifier when calling {% include api i="inter
  ```python
 await apps.launch_app("com.netflix.Netflix")
  ```
+
+To launch an app with a URL, pass the URL when calling {% include api i="interface.Apps.launch_app" %}
+
+ ```python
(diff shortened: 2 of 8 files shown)


postlund:
Thank you for submitting this and sorry for the _extremely_ long delay answering. Been busy with other things so pyatv had to be set aside for a bit. But I'm gonna try to catch up on things now.

Do you have any additional example URLs to test with? Assuming only one of the two (bundle id or URL) is needed, then I would actually prefer that we keep only one argument in `launch_app` and internally figure out which method to use. Does that seem feasible?

chapados:
Hi, thanks for the feedback. I think it is feasible to detect whether the parameter is a `bundleID` vs `URL` and act accordingly.  I will try to get this branch updated over the next few days.  Similar to @michalmo, I have used this with AppleTV+, Disney+ and HBOMax, but `nflx://` deep links don't work in my testing either and I haven't dug into it.  My main use case was launch content through the AppleTV+ app, using links derived from their web app, for example:

michalmo:
Is there anything I can do to help complete this?

michalmo:
Ok, the changes are simple, but I'll need some time to document the links.

postlund:
> Ok, the changes are simple, but I'll need some time to document the links.

No problem, we can wait until another release with this anyway.

michalmo:
@postlund I created #2024 based on this branch with the changes you requested and documentation.

Also, it seems that a recent update to Netflix.app has made Netflix URLs start working 🥳. I wasn't able to get working settings menu links. Even ones extracted from tvOS app strings don't work when used, I think there's some authentication scheme going on there.

postlund:
@michalmo It looks really good! 👍🎉I have merged it now. Is this a feature potentially something we will se officially in tvOS/iOS 17? Anyways, it's cool that pyatv makes use of it even before Apple has.

postlund:
Closing this now as #2024 has been merged!

michalmo:
@postlund my guess it's just part of the Handoff/Continuity protocol (which I understand works over Companion), using a URL to pass a deep link from one device to another. It also makes me think there must be a way to query the currently running app/link.

postlund:
@michalmo You are right, Handoff/Continuity uses Companion. Might be another frame type though (https://github.com/postlund/pyatv/blob/afe74a5f3257d56b3d7be10192dd55e3cd4202a2/pyatv/protocols/companion/connection.py#L20). Would be interesting to look at the traffic between an iOS device and macOS to see what happens. I guess the easiest way would be to attach lldb and breakpoint in an appropriate function, then dump the messages.

PR #1877, opened 2022-11-16
ability to ask AppleTV to open a user-provided URL

eeeebbbbrrrr:
This adds a `open_url=the://url/goes/here` command to pyatv.

I did some hacking to figure this one out.  I can't say there's any documentation around this.  It uses the "_launchApp" command, but specifies a "_urlS" instead of the bundle_id.

I can only assume AppleTV takes the url and asks each installed app, one-by-one, to open it.  If one of them agrees to open it, it will.  If none of them want it, it'll fail with:

```
pyatv.exceptions.ProtocolError: Command failed: Open URL failed
```

Some apps seem to have their own custom URL schemes, while others seem to respond to just general https://.  I'm not even going to attempt to document what I've learned about all that.

If there's anything else you might need from me for this PR, just ask.  I would like to see it merged so I don't have to maintain/install from my own fork.

### DIFF
--- a/pyatv/interface.py
+++ b/pyatv/interface.py
@@ -681,6 +681,10 @@ class Apps:
         """Launch an app based on bundle ID."""
         raise exceptions.NotSupportedError()
 
+    @feature(50, "OpenURL", "Open a URL.")
+    async def open_url(self, url: str) -> None:
+        """Open a URL."""
+        raise exceptions.NotSupportedError()
 
 class Metadata:
     """Base class for retrieving metadata from an Apple TV."""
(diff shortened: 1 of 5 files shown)


pyatv/interface.py
```diff
@@ -681,6 +681,10 @@ async def launch_app(self, bundle_id: str) -> None:
         """Launch an app based on bundle ID."""
         raise exceptions.NotSupportedError()
 
+    @feature(50, "OpenURL", "Open a URL.")
```
eeeebbbbrrrr, line 684: I picked the next highest number.  Dunno if that was correct or not?

chapados:
I opened [a similar PR to add openURL support](https://github.com/postlund/pyatv/pull/1829) a few months ago...

postlund:
Thank you for submitting 👍 As stated, @chapados made an implementation in #1829 as well and I think we can continue wtith that. I would prefer to just extend `launch_app` to also support URLs.

PR #1882, opened 2022-11-23
Artwork_save file naming.

AVITPL:
Added an extra parameter to the "artwork_save" function of the atvremote to be able to give the artwork file a specific name. If is not used, the file will still be called "artwork.png"


postlund:
Looks good to me! But I would like that you add instructions to the documentation (https://pyatv.dev/documentation/atvremote/#play-status) about this, so people now how to use it.

PR #1888, opened 2022-12-01
Redirecting to correct page

pranav767:
Hey, in the current getting started page , the current redirect to documentation and development is giving a 404 error as the redirect link was incorrect.


postlund:
Thanks! 👍

PR #1908, opened 2023-01-30
Meaningful paired device names for AirPlay and Companion protocols

michalmo:
This PR adds support for settings meaningful device names during pairing of AirPlay and Companion protocols.

Currently the device name visible in Settings -> Remotes and Devices -> Remote App and Devices is "Living Room" for paired Companion protocol, and "Unknown" for AirPlay.

With this PR the default will be "pyatv", and can be changed with the --remote-name option of `atvremote`.  A corresponding change can be made to Home Assistant (https://github.com/home-assistant/core/blob/8f74bff354374357f25f62469980441e71e01f8f/homeassistant/components/apple_tv/config_flow.py#L400) once this is merged.

I have also noticed that the HAP credentials for AirPlay and Companion protocols are interchangeable, so the Home Assistant config flow can be simplified to ask for a PIN only once.


postlund:
Looks good to me! I would possibly prefer att test for the name by extracting it in the fake device, but it's not really that important that I might let it slide this time. I'll merge when checks pass.

PR #1905, opened 2023-01-24
Add support for companion protocol text input

michalmo:
This PR adds support for interacting with the text in the on screen keyboard using RTI text operations over the Companion protocol. Supported operations are reading the current text, appending text, clearing the text, and replacing the text.

Resolves #1716

PR #1912, opened 2023-02-04
Fix companion proxy

michalmo:
This PR gets the Companion protocol support of atvproxy to a working state.

Contrary to what was written in the documentation, i've discovered that the Apple TV ignores the signatures in the `_systemInfo` request. However, both ends of the connection do track a persistent unique identifier (the `_i` property of the request and response) and will terminate old connections with the same identifier.

The events that caused the proxy to fail before this PR were:
1.  The proxy would connect to the Apple TV with the same identifier as the direct connection from the iOS device.
2. The Apple TV would terminate the direct connection from the iOS device.
3. The iOS device would attempt to reconnect with the same identifier, after being unexpectedly disconnected.
4. The Apple TV would disconnect the now older proxy connection.
By coincidence the disconnection appeared in the logs close to the lines about signatures, which I assume lead to this misunderstanding.

Other issues were related to incorrect OPACK re-encoding, incorrect encryption when receiving NoOP frames from the client device, and name and identifier collisions, which apparently cause the iOS device to sometimes connect directly to the target device bypassing the proxy. I assume the iOS device keeps a mapping of identifiers to IP addresses.

I have also updated the published mdns properties, to expose some values from the proxied device, and replace others to avoid collisions that prevented the proxy from working consistently. The values used were from an Xcode emulated Apple TV device, since the previous ones were not working consistently for me.

Occasionally iOS devices will fail to complete the connection ("Connecting" visible in the remote app despite seeing initialization traffic in the proxy). In such situations rebooting the iOS device, restarting the proxy, restarting the Apple TV Remote app (in newer iOS versions), or a combination of these gets it working again. 

But it's working well enough to share, and allowed me to complete [a PR with support for keyboard input](https://github.com/postlund/pyatv/pull/1905).


pyatv/helpers.py
```diff
@@ -69,6 +70,11 @@ def get_unique_id(  # pylint: disable=too-many-return-statements
         return properties.get("UniqueIdentifier")
     if service_type == AIRPLAY_SERVICE:
         return properties.get("deviceid")
+    if service_type == COMPANION_SERVICE:
+        # Apple TV devices on tvOS 16 (maybe earlier) have a static rpMRtID
+        # identifier. For other devices use rpBA which periodically changes but is
+        # still useful.
+        return properties.get("rpmrtid") or properties.get("rpba")
```
postlund, line 77: the identifer here must be unique, so it is ok to keep `rpmrtid` but not `rpba`.

pyatv/protocols/companion/opack.py
```diff
@@ -97,7 +97,17 @@ def _pack(data, object_list):
 
     # Re-use if in object list, otherwise add it to list
     if packed_bytes in object_list:
-        packed_bytes = bytes([0xA0 + object_list.index(packed_bytes)])
+        object_index = object_list.index(packed_bytes)
```
postlund, line 100: Great that you added these! Some tests are needed, at least for `object_index <= 0xFF` and possibly 0xFF (the other ones might be hard without using a lot of memory).

postlund (changes requested):
Some minor notes, but otherwise really great work! Tried it out briefly and seems to do the job a lot better. Do you know what is required to make the Shortcut app pick the device up? It doesn't seem to do that (and I haven't succeeded with that before AFAIK).

pyatv/helpers.py
michalmo: Since only Apple TVs report `rpmrtid` (which I assume is an identifier for the remote app), I added `rpba` to be able to more easily run scripts against the proxy connected to a HomePod (otherwise it has no identifier). The `rpba` seams to be unique although it isn't permanent. I'll remove it from the PR though.

michalmo: Done.

pyatv/protocols/companion/opack.py
michalmo: Done.

postlund (approved):

postlund:
> > Do you know what is required to make the Shortcut app pick the device up? It doesn't seem to do that (and I haven't succeeded with that before AFAIK).
> 
> From my testing the proxy appears in the remote app always when it's advertised, but in the shortcuts app only if it's able to successfully connect. Also not all commands in the shortcuts app show the proxy. The Apple TV remote ones do, the music player ones (which I assume use the AirPlay protocol) don't.
> 
> If something isn't showing up, I've found that restarting things consistently solves it as described in the PR's description ("Occasionally iOS devices will fail to complete the connection ...").

Right, I will wiggle it a bit and see if it works better. Thanks for mentioning that! 👍

michalmo:
I'm not able to reproduce this [pylint error](https://github.com/postlund/pyatv/actions/runs/4742921556/jobs/8421794691?pr=1912#step:10:43) in either Python 3.9.16 or 3.8.16. @postlund is this just a CI issue?

postlund:
I will try to reproduce it later today.

postlund:
Hmm, I can't reproduce it myself either. Looks like pylint gets confused about the types for some reason. Since `self.listener` can be `None`, there should be an if-clause checking for that. Maybe you can add that and see what happens (at this point it looks like a pylint bug to me)?

michalmo:
[This change](https://github.com/postlund/pyatv/pull/1912/commits/b7b0eb37169fcf71645bcf6b49ac60289bc23f2b) resolves the pylint issue.

pyatv/protocols/companion/connection.py
```diff
@@ -90,12 +90,18 @@ def enable_encryption(self, output_key: bytes, input_key: bytes) -> None:
         """Enable encryption with the specified keys."""
         self._chacha = chacha20.Chacha20Cipher(output_key, input_key, nonce_length=12)
 
+    def set_listener(self, listener: CompanionConnectionListener) -> None:
```
postlund, line 93: Since `set_listener` should be used from now on, please change the variable name to `_listener` to show this intent. It will require additional change here as well:

https://github.com/postlund/pyatv/blob/f8fbbfe5373cd6554bb5e8d008fbbbe5719bf5c0/pyatv/protocols/companion/protocol.py#L85

Maybe some tests are affected as well, not sure.

michalmo: It's not really my intention to change the API, which is used also in [`MrpAppleTVProxy`](https://github.com/postlund/pyatv/blob/f8fbbfe5373cd6554bb5e8d008fbbbe5719bf5c0/pyatv/scripts/atvproxy.py#L60) and possibly in user code somewhere, just to work around the issue that pylint appears to be incorrectly inferring the type based on ```self.*.listener = self```.

How about [this alternative fix](https://github.com/michalmo/pyatv/commit/b501c91dfb5cb14626bdcaef96a8fbea401452df)? this way the change is localized to `CompanionAppleTVProxy`.

postlund: This is an internal API, so no one should be using it outside of pyatv. So it's totally fine to change it, I just want to make sure we are consistent and use one way of doing things.

michalmo: Alright, done https://github.com/postlund/pyatv/pull/1912/commits/6be6c7489bc73466929789d2b7e1a2c3646f638c

postlund:
Looks good now 👍

Issue #1941, opened 2023-04-01
OS 16.4 HomePod Introduces stream file failure


Ghawken:
### Describe the bug

Disconnection of Homepod when attempt to stream_file 
(previously working - Note play_url was not)

### Error log

```log
01-04-2023 12:24:04.726	DEBUG	pyatv.core.protocol.dispatch:protocol.py:115:	Dispatching message with type 46 to <bound method PlayerStateManager._handle_set_now_playing_client of <pyatv.protocols.mrp.player_state.PlayerStateManager object at 0x11860ca90>>
01-04-2023 12:24:04.727	DEBUG	pyatv.auth.hap_channel.log_binary:__init__.py:56:	Sending data (Encrypted=200025d1965b57f1d9185d3aecdebe729169acec5821268ddbcd7cc1d8a9e11a401884ed406a37f77dd247cfafbec21743a2)
01-04-2023 12:24:04.727	DEBUG	pyatv.protocols.mrp.player_state._handle_set_now_playing_client:player_state.py:268:	Active client is now 
01-04-2023 12:24:04.727	DEBUG	pyatv.core.protocol.dispatch:protocol.py:115:	Dispatching message with type UpdatedState.Playing to <bound method FacadePower._playing_changed of <pyatv.core.facade.FacadePower object at 0x11860f610>>
01-04-2023 12:24:04.727	DEBUG	Plugin.powerstate_update:plugin.py:270:	HomePod -new- Kitchen Powerstate update Old PowerState.On and new PowerState.Off
01-04-2023 12:24:04.729	DEBUG	Plugin.playstatus_update:plugin.py:328:
```

### How to reproduce the bug?

Stream_file to OS 16.4 Homepod

1/2 time works

When fails (below) power_state appears to update from Off to On, just prior:
```
nfoMessage] {
  uniqueIdentifier: "AF27F100-B793-4254-A8DC-D01495803E26"
  name: "Kitchen"
  localizedModelName: "HomePod"
  systemBuildVersion: "20L497"
  applicationBundleIdentifier: "com.apple.mediaremoted"
  protocolVersion: 1
  lastSupportedMessageType: 130
[Works.txt](https://github.com/postlund/pyatv/files/11128480/Works.txt)

[Fails.txt](https://github.com/postlund/pyatv/files/11128479/Fails.txt)

  supportsSystemPairing: true
  allowsPairing: true
  systemMediaApplication: "com.apple.Music"
  supportsACL: true
[...]

Issue #1886, opened 2022-11-28
Pause labelled as Play in interface.py


Ghawken:
### Describe the bug

Hi

Simple fix - in interface.py

Pause button (line 295 interface.py) is labelled as “Press key play.”
Causing this to be command summary when pulled.

Thanks

### Error log

```log
<place log here>
```

### How to reproduce the bug?

as above

### What is expected behavior?

as above

### Operating System

macOS

### Python

3.10

### pyatv

Latest

### Device

all

### Additional context

nil

PR #1942, opened 2023-04-01
Fix for OS16.4 Stream file Timeout error

Ghawken:
Seems to have been a number of OS16.4 changes which slows down the handling, leading to timeout of streaming file

Increasing timeout to 10 seconds resolves.

Fix here for mislabelling of play/pause

Closes:
#1941 
#1886

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # CHANGES
 
+Fix http timeout error, noted with OS16.4 upgrade
+Fix Pause/Play button mis-labelled
+
 ## 0.10.3 Chica (2022-08-02)
 
 This is another minor release:
--- a/pyatv/const.py
+++ b/pyatv/const.py
@@ -5,7 +5,7 @@ from enum import Enum
 
 MAJOR_VERSION = "0"
 MINOR_VERSION = "10"
-PATCH_VERSION = "3"
+PATCH_VERSION = "4"
 __short_version__ = f"{MAJOR_VERSION}.{MINOR_VERSION}"
 __version__ = f"{__short_version__}.{PATCH_VERSION}"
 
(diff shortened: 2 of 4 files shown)


CHANGES.md
```diff
@@ -1,5 +1,8 @@
 # CHANGES
 
+Fix http timeout error, noted with OS16.4 upgrade
```
postlund, line 3: `CHANGES.md` are updated when I make a new release, so you can remove this.

pyatv/const.py
```diff
@@ -5,7 +5,7 @@
 
 MAJOR_VERSION = "0"
 MINOR_VERSION = "10"
-PATCH_VERSION = "3"
+PATCH_VERSION = "4"
```
postlund, line 8: Same here, you should not update version manually.

postlund (changes requested):
I can confirm that the required to set up the stream takes longer now for some reason. Seems to be around six seconds:

```
2023-04-08 09:46:22 DEBUG [pyatv.support.http]: Sending RTSP/1.0 message: b'POST /feedback RTSP/1.0\r\nUser-Agent: AirPlay/540.31\r\nCSeq: 3\r\nDACP-ID: 54B6110849A4851B\r\nActive-Remote: 2690020292\r\nClient-Instance: 54B6110849A4851B\r\n\r\n'
2023-04-08 09:46:28 DEBUG [pyatv.support.http]: Received: b'RTSP/1.0 200 OK\r\nDate: Sat, 08 Apr 2023 07:52:40 GMT\r\nContent-Length: 0\r\nServer: AirTunes/680.23.1\r\nCSeq: 1\r\n\r\n'
```
So I think 10 seconds is fine.

pyatv/const.py
Ghawken: Sorry!  think I had to here - to double check had pulled fork version.
But feel free to please cancel and update you end (away for a bit)

postlund:
Fixed by #1959.

Issue #1848, opened 2022-09-29
Artwork is always a black image on tvOS 16 on Homepod Mini only


NebzHB:
### Describe the bug

when i run artwork_save, i always receive a black image on homepod mini (homepod and atv is working fine !) (probably since tvOS 16...)

if i get the id and download the artwork it's ok... i've tried with 1200 x 1200 to avoid resizing maybe... but doesn't work

### Error log

see attachment (too long) (at the end there is a artwork_id and a wget of the url)

### How to reproduce the bug?

doing the commands on homepod mini

### What is expected behavior?

receiving artwork.png with the correct artwork (works fine on homepod and atv !!)

### Operating System

Linux debian 11

### Python

3.9

### pyatv

0.10.3

### Device

Homepod mini tvOS 16.0

### Additional context

no

NebzHB:
[pyatv.log](https://github.com/postlund/pyatv/files/9673143/pyatv.log)

Clarus-Moof:
Same here. When I added a new HomePod mini to HA, the artwork was visible for a short time, but after a while it also turned black.

michalmo:
@postlund #1958 has been merged and released, so this can be closed. I didn't correctly link that PR to this issue.

postlund:
@michalmo Got it, thanks! 👍

PR #1958, opened 2023-04-14
Fix missing artwork

michalmo:
Missing artwork has been reported at #1848 and https://github.com/home-assistant/core/issues/81185https://github.com/home-assistant/core/issues/81185.

My own investigation shows that this is not an issue isolated to HomePods. tvOS seems to only fetch the artwork locally for select media sources - it does for video media from streaming apps, but does not for music media from Apple Music (instead returns links to itunes artwork hosted by Apple).

This PR will fetch the artwork from the `artworkIdentifier` or `artworkURL` fields of the metadata if available.

Home Assistant supports returning artwork urls instead of artwork data, but that would require an API change in pyatv so I wend with this approach.

### DIFF
--- a/pyatv/protocols/mrp/__init__.py
+++ b/pyatv/protocols/mrp/__init__.py
@@ -6,6 +6,9 @@ import logging
 import math
 import re
 from typing import Any, Dict, Generator, List, Mapping, Optional, Set, Tuple
+from urllib.parse import urlparse
+
+import aiohttp
 
 from pyatv import exceptions
 from pyatv.auth.hap_srp import SRPAuthHandler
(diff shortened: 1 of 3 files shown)


pyatv/protocols/mrp/__init__.py
```diff
@@ -502,6 +505,58 @@ async def artwork(
...
+                url = None
+            url_parts = urlparse(url)
+            if url_parts.scheme and url_parts.netloc:
+                urls.append(url)
+
+        if metadata.HasField("artworkURL"):
+            # artworkURL has fixed size and format, use it as a fallback
+            urls.append(metadata.artworkURL)
+
+        async with aiohttp.ClientSession() as session:
```
postlund, line 541: You should not create a new session here, but rather re-use the one in the `pyatv.core.Core` instance (passed when instantiating the protocol). That session is re-used everywhere in pyatv and is usually provided by the user.

postlund (changes requested):
I have not tried it myself, but it looks solid to me. This is something I would like a test for though. It should be fairly simple to set up a small web server using `pyatv.support.http.BasicHttpServer`. Just provide a handler that returns a fixed value, make the fake MRP device include a URL to said web server and verify the fetch artwork matches what is returned by the handler.

postlund (approved):
Looks good! 👍

haywiremk:
Any update on what version of HA(https://www.home-assistant.io/integrations/apple_tv) this fix may get merged into?

michalmo:
There will need to be a new release that includes this PR, and then HA will be updated to point at this latest release. However it may take some more time, the project owner seems to be busy at the moment.

PR #1975, opened 2023-04-21
Add listener interface to Audio for volume updates

### LINKED ISSUE #1798: push_updates with volume modifications (from lib or from other means) for homepods
### What feature would you like?

receiving {"volume":20.5} in atvscript push_updates when the volume has been modified from any means (lib, +/- buttons on top, airplay, ...) on HomePods at least (i know volume management on apple tv is strange).

could be a part of the playing json or separated like power_state, as you wish

### Describe the solution you'd like

each time the volume is changed, a line should appear in push_updates

### Any other information to share?

no

michalmo:
Adds a listener interface to `Audio`, to enable push updates for volume changes as recommended in https://github.com/postlund/pyatv/issues/1798#issuecomment-1188276550

Resolves #1798

### DIFF
--- a/docs/api/pyatv/interface.html
+++ b/docs/api/pyatv/interface.html
@@ -73,6 +73,12 @@ link_group: api
 </ul>
 </li>
 <li>
+<h4><code><a title="pyatv.interface.AudioListener" href="#pyatv.interface.AudioListener">AudioListener</a></code></h4>
+<ul class="">
+<li><code><a title="pyatv.interface.AudioListener.volume_update" href="#pyatv.interface.AudioListener.volume_update">volume_update</a></code></li>
+</ul>
+</li>
+<li>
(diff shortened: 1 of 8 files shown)


postlund:
I have not tried it, but it looks really good and I have confidence in you 👍  Will merge when checks pass.

michalmo:
Thanks, I'll make a corresponding PR for Home Assistant after.

This ROAP test failure doesn't seem related, is there some timeout happening?

postlund:
> Thanks, I'll make a corresponding PR for Home Assistant after.
> 
> This ROAP test failure doesn't seem related, is there some timeout happening?

Great!

Nah, that's a flaky test I should take care of (some day).

postlund:
I can do a 0.11.1 release tomorrow btw with this change so you can actually get it in to Home Assistant.

Issue #2005, opened 2023-05-24
HomePod reporting volume of grouped AirPlay device instead of self


michalmo:
### Describe the bug

While working on adding support for push volume updates to HA (enabled by #1975), I came across a problem when testing an example automation.

Pyatv will report and set the volume of the device that most recently joined the AirPlay group (more specifically the device whose UID was in the most recent `VOLUME_CONTROL_CAPABILITIES_DID_CHANGE_MESSAGE`).

### Error log

N/A

### How to reproduce the bug?

1. With the current master branch, start `atvscript -s <IP of HomePod> push_updates` in a terminal.
2. From an iOS device control center, switch to controlling the HomePod and verify that changing volume generates output from atvscript.
3. Group a second output device with the HomePod from the control center.
4. Changing volume of the HomePod will no longer generate output from atvscript, instead changing the volume of the second device will.

### What is expected behavior?

Since I can connect to another AirPlay device to get it's volume, I would expect the HomePod to always report it's own volume level. In fact the automation I wanted to test (sync HomePod volume to other devices) requires being able to always get the HomePod's own volume.

### Operating System

macOS/Home Assistant OS

### Python

3.10

### pyatv

master branch

### Device

HomePod tvOS 16.5

### Additional context

This prevents push volume updates in Home Assistant from being useful.

Issue #1838, opened 2022-09-10
volume and set_volume doesn't work


iPhilBln:
### Describe the bug

The commands volume and set_volume doesn't work for me. If I use the companion or airplay protocol no errors show up but nothing happen. 

* atvremote --id DC56E74FE734 --companion-credentials c66a968278b2d609.... set_volume=10.0 doesn't change the volume
*  atvremote --id DC56E74FE734 --companion-credentials c66a968278b2d609.... volume always return 0.0

*  atvremote --id DC56E74FE734 --companion-credentials c66a968278b2d609.... volume_up works 
*  atvremote --id DC56E74FE734 --companion-credentials c66a968278b2d609.... volume_down works 

### Error log

[pyatv.log](https://github.com/postlund/pyatv/files/9540838/pyatv.log)

### How to reproduce the bug?

 atvremote --id DC56E74FE734 --companion-credentials c66a968278b2d609c3f17.... set_volume=10.0 --debug > pyatv.log
 atvremote --id DC56E74FE734 --companion-credentials c66a968278b2d609c3f17....  volume --debug >> pyatv.log

### What is expected behavior?

setup the volume and get the current volume

### Operating System

Linux Debian Bullseye

### Python

3.9

### pyatv

0.10.3

### Device

Apple TV 4, tvOS 15.6, Apple TV 4K, tvOS 15.6

### Additional context

nothing to say

postlund:
This should be fixed in `master`, but it's not released yet. Would be great if you could verify that!

iPhilBln:
Ok I will test it tomorrow. 👍

iPhilBln:
> This should be fixed in `master`, but it's not released yet. Would be great if you could verify that!

I am a little bit confused, which branch i should test? I am running the latest master branch v. 0.10.3 and there it doesn't work.

earlywaff:
@postlund Really no volume setting works on any of the branches:)

JonLaliberte:
Hm. So it looks like @postlund pushed a change to try to fix this, but that was back in August and it was released as 0.10.3 (before this issue was created). I wonder if that change ended up breaking things in some circumstances, or if maybe it didn't fix the original problem in all circumstances?

[...]

PR #2006, opened 2023-05-24
Make sure were sending and receiving volume messages for our device

michalmo:
… even if playing to other AirPlay devices in a group.

Fixes: #2005

Instead of updating `self._output_device_id` when we get a `VOLUME_CONTROL_CAPABILITIES_DID_CHANGE_MESSAGE` (which may be the UID of another grouped AirPlay device), just always use the one from device info which we get when the protocol starts.

### DIFF
--- a/pyatv/protocols/mrp/__init__.py
+++ b/pyatv/protocols/mrp/__init__.py
@@ -726,16 +726,13 @@ class MrpAudio(Audio):
         self.protocol: MrpProtocol = protocol
         self.state_dispatcher = state_dispatcher
         self._volume_controls_available: bool = False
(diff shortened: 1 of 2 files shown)


postlund:
This looks reasonable to me, will merge it when it passes. I didn't find the time to make a release yesterday, but I can make it now pretty much. Is there anything else you want to have merged before the release?

michalmo:
Just this one will allow me to make the changes to HA.

#1905 is still open when you have some time, but that can be later.

JonLaliberte:
I'm wondering if this would also resolve #1838 ?

michalmo:
I'll answer in that issue, since there's quite a bit of detail.

PR #2013, opened 2023-05-26
Add Apple TV 4k (gen3) and new tvOS versions

Shraymonks:
Add Apple TV 4k (gen3):
https://www.theiphonewiki.com/wiki/J255AP

Add new tvOS versions:
https://en.wikipedia.org/wiki/TvOS#tvOS_16


postlund:
Great, thanks for updating! 👍

PR #2031, opened 2023-06-05
Keyboard focus events and documentation

michalmo:
This PR adds a listener interface for updating the focus state of the virtual keyboard, and documentation for the Keyboard interface as discussed in comments to #1905.

I've opted to not create a listener for text content, or to pass the current text as part of the focus state update since my testing has found this unreliable. The device only sends text updates for changes made through on screen clicks, not made over the API, and only appears to have started doing so in a patch release that occurred after developing the initial PR. Using text_get gets the actual content always by resetting the text input session before each request.

### DIFF
--- a/pyatv/interface.py
+++ b/pyatv/interface.py
@@ -1122,7 +1122,7 @@ class KeyboardListener(ABC):  # pylint: disable=too-few-public-methods
         self, old_state: const.KeyboardFocusState, new_state: const.KeyboardFocusState
     ):
         """Keyboard focus state was updated."""
-        raise NotImplementedError()
+        raise exceptions.NotSupportedError()
 
 
 class Keyboard(ABC, StateProducer):
--- a/pyatv/interface.py
+++ b/pyatv/interface.py
@@ -1114,8 +1114,29 @@ class Audio(ABC, StateProducer):
         raise exceptions.NotSupportedError()
 
 
-class Keyboard:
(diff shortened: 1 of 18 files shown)


IsakTheHacker (approved):

pyatv/interface.py
```diff
@@ -1114,8 +1114,29 @@ async def volume_down(self) -> None:
...
-    """Base class for keyboard handling."""
+class KeyboardListener(ABC):  # pylint: disable=too-few-public-methods
+    """Listener interface for keyboard updates."""
+
+    @abstractmethod
+    def focusstate_update(
+        self, old_state: const.KeyboardFocusState, new_state: const.KeyboardFocusState
+    ):
+        """Keyboard focus state was updated."""
+        raise NotImplementedError()
```
postlund, line 1125:
```suggestion
        raise exceptions.NotSupportedError()
```

postlund (changes requested):
Looks good to me (with one minor exception comment)! 👍

IsakTheHacker:
@postlund When will you release a new version with this functionality included? I really need the focus events to finish my app

postlund:
@IsakTheHacker I can make a release later tonight.

IsakTheHacker:
That's awesome! Thank you so much

postlund:
@IsakTheHacker Release is cooking in #2039

PR #2033, opened 2023-06-07
Fix intermittent companion proxy connection issues

michalmo:
Companion protocol appears to bee sensitive to the byte-size of opack encoded integers. Previously proxy connection could sometimes fail if the size of the response "_x" field did not match the size of the request "_x" field.

This PR preserves  integer sizes when decoding/re-encoding with opack


postlund:
Peculiar behavior on Apples part, but OK!

PR #2044, opened 2023-06-12
Control AirPlay output devices

### LINKED ISSUE #1939: How can I connect speakers with Apple TV via ssh?
### What do you need help with?

Hi there! I have RPi with installed pyatv so now I can connect to my ATV HD via ssh, i.e. I can switch it on or off, pausing it or stream some file/url etc. But I still can't realize how to connect some airplay2 speakers with Apple TV via ssh. Could you explain it somehow please?

michalmo:
Adds support for controlling AirPlay output devices over MRP, and the necessary elements to integrate this feature into Home Assistant.

Resolves #1939.


pyatv/interface.py
```diff
@@ -912,6 +912,7 @@ class DeviceInfo:
     MODEL = "model"
     RAW_MODEL = "raw_model"
     MAC = "mac"
+    AIRPLAY_IDENTIFIER = "airplay_id"
```
postlund, line 915: I would prefer to rename this to `output_device` (and correspondingly change "AIRPLAY") to de-couple from AirPlay as a technology as much as possible. Theoretically we could get another technology in the future.

pyatv/interface.py
```diff
@@ -1059,6 +1068,35 @@ def in_state(
...
+        return f"Device: {self.name} ({self.identifier})"
+
+    def __eq__(self, other) -> bool:
+        """Return self==other."""
+        if isinstance(other, OutputDevice):
+            return self.name == other.name and self.identifier == other.identifier
+        return False
+
+
 class AudioListener(ABC):  # pylint: disable=too-few-public-methods
```
postlund, line 1100: I think this supression can be removed now.

pyatv/protocols/mrp/__init__.py
```diff
@@ -823,6 +837,41 @@ async def volume_down(self) -> None:
...
+
+    @property
+    def output_devices(self) -> List[OutputDevice]:
+        """Return current list of output device IDs."""
+        return self._output_devices
+
+    async def add_output_devices(self, *devices: List[str]) -> None:
+        """Add output devices."""
+        await self.protocol.send(messages.add_output_devices(*devices))
+        with suppress(asyncio.TimeoutError):
```
postlund, line 860: Are these errors frequent? Generally I prefer if errors bobble up so users are aware of an issues rather than hiding them.

postlund (changes requested):
Minor things, looks good and well tested. Great work!

pyatv/protocols/mrp/__init__.py
michalmo: This will happen any time we don't get a DEVICE_INFO_UPDATE_MESSAGE after sending the change request, basically any time the request doesn't cause an actual change. It would be harder to always keep track of when a change will or will not happen.

I opted to suppress it since users seem to find it confusing (see e.g.: #1838; it's not the command that timed out, but the response).

postlund: Ok, that is good enough then!

perfrankling:
Is this feature implemented in Home Assistant?

PR #2047, opened 2023-06-13
add FeatureName.TextFocusState to supported features for Companion

michalmo:
I forgot to add this in #2031.


postlund:
Yeah, that's right. I think there should be some modifications for these constants regarding the Features interface. I probably missed that in the previous review, can you take a look at that?

michalmo:
I'm not sure I understand. It looks like all the other protocols just fall back to Unsupported or Unavailable for these new feature constants, and this is the only one I forgot to add for Companion.

Can you explain what you have in mind?

postlund:
> I'm not sure I understand. It looks like all the other protocols just fall back to Unsupported or Unavailable for these new feature constants, and this is the only one I forgot to add for Companion.
> 
> Can you explain what you have in mind?

All features we add needs to have some kind of state support. In case of companion we always know that the feature exists and are available (with media flags being an exception, since we get updates for play, pause, etc.) so we should make sure we return "available" in that case. Simply add feature to this list (you should probably add for output devices as well):

https://github.com/postlund/pyatv/blob/254694c9b9a304bc5161437d1c99952e0a782755/pyatv/protocols/companion/__init__.py#L107

And update corresponding test to verify it. Should be a list in the test as well.

michalmo:
OK. I added the features as supported previously (except FeatureName.TextFocusState). Just added a commit with tests for all the new ones too.

PR #2049, opened 2023-06-14
Volume control improvements

michalmo:
This PR changes MRP volume control to be aware of the device `volumeCapabilities` (`Absolute` or `Relative`).

Should fix https://github.com/home-assistant/core/issues/76117 for HomePods which only support `Absolute` and do not respond to HIP volume up/down commands.

For Apple TV it should set the Volume and SetVolume features as Unavailable when the device reports `Relative`, so it should be easier to explain behaviour in tickets such as #1838.
Home Assistant will still say it supports the volume features in this case (https://github.com/home-assistant/core/blob/dev/homeassistant/components/apple_tv/media_player.py#L125-L128) and show the volume slider, but now pyatv will not throw a TimeoutError.

### DIFF
--- a/pyatv/protocols/mrp/__init__.py
+++ b/pyatv/protocols/mrp/__init__.py
@@ -731,6 +731,8 @@ class MrpAudio(Audio):
         self.protocol: MrpProtocol = protocol
         self.state_dispatcher = state_dispatcher
         self._volume_controls_available: bool = False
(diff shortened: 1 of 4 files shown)


postlund:
Looks, great! Will definitely help some users having problems with volume changes in the past 👍  It's also great that Apple has managed to make the simplest feature so utterly complex. Really great.

Issue #2064, opened 2023-06-19
play_url not working


HennieLP:
### Describe the bug

if i try ```atv.stream.play_url("http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4")```
the apple tv goes plack and shows an Loading symbol but after 4 second it goes back to the home screen

### How to reproduce the bug?

use  ```atv.stream.play_url("http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4")```

### What is expected behavior?

it plays the video on the apple tv

### Operating System

windows

### Python

Other

### pyatv

0.13.0

### Device

Apple TV 4K (gen2) TVos 16.3.1

### Additional context

-

postlund:
This is a known open issue that I more or less given up on. After some digging and trial an error though, I have managed to restore most of the functionality via AirPlay 2 (supporting newer devices). My main issue right know is that the video starts in paused state and I can't seem to figure out how to fix that. Will have to continue testing. There's a lot of cleanup to do as well, so I don't expect to have a fix for this for a while. Probably after summer or something like that.

postlund:
I figured out how to get playback to start automatically, so I think I can fix this.

postlund:
Ah, it's probably because I'm not sending any response on the event channel.

postlund:
Well yep, seems like that was it. Rookie mistake that I have made before...

postlund:
@HennieLP Just pushed #2079 which should fix this, can you verify this?

postlund:
It is currently not possible to stop the playback but I will fix that before merging the PR.

oxplot:
@postlund I can confirm that not *only* does it work, but it works better than before. Specifically, running `atremote play_url=` no longer requires the TV to be on to work properly. It turns on the TV and starts playing automatically.

In other words, tremendous success and I'm very happy.

postlund:
@oxplot That's great! 👍 I'm gonna try to fix `stop` and then I'll merge it.

seidtgeist:
@postlund Absolute legend! I had "Require password" enabled on my Apple TV 4K Gen1 which tripped up the RTSP authentication, but once it's disabled everything is working perfectly. Thank you!

postlund:
@seidtgeist 😎 Yeah, password is not supported for AirPlay yet but should probably be fixed at some point. It's already implemented for RAOP, so it should be fairly easy to just port that over.

[...]

PR #2079, opened 2023-06-30
airplay: Fix play_url with newer tvOS versions

postlund:
The play_url method has been broken for a very long time as I believed
Apple wasn't maintaining the underlying functionality anymore. They do,
but an AirPlay session must be established for it to work now. So I
have added that and verified with an Apple TV 3 and 4K running
tvOS 14.5.

Relatest to #2064

<a href="https://gitpod.io/#https://github.com/postlund/pyatv/pull/2079"></a>

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,15 +8,14 @@ A client library for Apple TV and AirPlay devices
 [![codecov](https://codecov.io/gh/postlund/pyatv/branch/master/graph/badge.svg)](https://codecov.io/gh/postlund/pyatv)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 [![PyPi Package](https://badge.fury.io/py/pyatv.svg)](https://badge.fury.io/py/pyatv)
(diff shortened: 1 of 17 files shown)


IsakTheHacker (approved):

Ghawken:
Testing this on my non-HA use of your great and getting greater, pyatv library.

Play_url works on my testing (>OS 16.5)
Apple4k gen 3

Works with:
https://gist.github.com/jsturgis/3b19447b304616f18657
Pretty much mp4s here

Get an NSOSDomainError when trying to playback camera streams from BlueIris, or camera direct.
Either http://temp.m fails and so does raw camera stream.
Presume  this is because of nature of stream, encoders used etc.

postlund:
Thanks for the detailed testing @Ghawken (and sorry for the late response). I honestly don't know what AirPlay is supposed to support in this particular use case, so I can't really tell what is expected to work or not. Figuring that out would be good. Feel free to open a separate issue and attach any findings you make there to keep track of it, maybe we can make some hacks to extend support over time.

PR #2083, opened 2023-07-02
Rework zeroconf for unicast scanning

bdraco:
I've tested this on both my large production setups without issues with many airplay, home pods, and apple tv devices.  I think it will solve https://github.com/home-assistant/core/issues/80215

It would be great if someone with an older apple tv could test this.

~~needs https://github.com/python-zeroconf/python-zeroconf/pull/1192~~

~~needs https://github.com/home-assistant/core/pull/95714~~

fixes https://github.com/home-assistant/core/issues/80215


pyatv/core/scan.py
```diff
@@ -392,9 +391,16 @@ async def _lookup_services_and_models(
     ) -> Tuple[Dict[str, List[AsyncServiceInfo]], Dict[str, str]]:
         """Lookup services and aggregate them by address."""
         infos = self._build_service_info_queries()
-        await asyncio.gather(
-            *[info.async_request(self.zeroconf, zc_timeout) for info in infos]
-        )
+        requests: List[Awaitable[bool]] = []
+        multicast_requests = [
+            info.async_request(self.zeroconf, zc_timeout) for info in infos
+        ]
```
bdraco, line 397: I think we can take this out if there are `hosts` in `self.hosts` but leaving it in for testing

bdraco:
We should try load_from_cache first like we do in aiohomekit

Screenshot 2023-07-29 at 6 18 00 AM

pyatv/core/scan.py
```diff
@@ -391,10 +390,26 @@ async def _lookup_services_and_models(
...
+                        [
+                            info.async_request(self.zeroconf, zc_timeout, host)
+                            for info in infos
+                        ]
+                    )
+                continue
+            # Multicast requests
+            requests.append(info.async_request(self.zeroconf, zc_timeout))
+        if requests:
+            await asyncio.gather(*requests)
```
bdraco, line 412: This block will do network traffic

bdraco:
the behavior should be the same exact we don't create hundreds of tasks now if they are just going to return right away because its in the cache
Screenshot 2023-07-29 at 6 34 45 AM

bdraco:
Screenshot 2023-07-29 at 6 35 40 AM

and if its not in the cache we do create the tasks

bdraco:
I'm pretty sure this will solve https://github.com/home-assistant/core/issues/80215 and it doesn't regress any of my production set ups.  It would be nice to get a test from someone else though

pyatv/core/scan.py
```diff
@@ -322,11 +322,10 @@ def _name_without_type(name: str, type_: str) -> str:
...
-def _first_non_link_local_or_non_v6_address(addresses: List[bytes]) -> Optional[str]:
-    """Return the first non ipv6 or non-link local ipv4 address."""
-    for address in addresses:
-        ip_addr = ip_address(address)
-        if not (ip_addr.is_link_local or ip_addr.version == 6):
+def _first_non_link_local_address(addresses: List[IPv4Address]) -> Optional[str]:
+    """Return the first non-link local ipv4 address."""
+    for ip_addr in addresses:
+        if not ip_addr.is_link_local:
             return str(ip_addr)
```
bdraco, line 329: zeroconf now has a strong guarantee on order so we can remove some of the code here

bdraco:
Since this path is used for HA we could do a release, and bump HA at the top of the month to get a long feedback period

postlund:
Just let me know when you are satisfied here and want this merged. I can probably make a release if needed.

bdraco:
I think the code is fine. 

I'm internally debating if a test that patches to the code to make sure we do the correct zeroconf calls makes sense or if there is a better way to do it. I think patching is probably the only realistic option since otherwise we end up with an unreliable test that needs specific network conditions

bdraco:
even with checking the cache first, there are still too many requests (this is an existing problem)

https://github.com/python-zeroconf/python-zeroconf/pull/1205

bdraco:
fixed that issue. will look at the tests a bit more later

postlund:
I suggest patching for tests as well. That's easier and more reliable.

bdraco:
Happy with the performance now. Will bump the zeroconf version and add the tests

bdraco:
@postlund should be good to go now.

postlund:
@bdraco Great! Heading to bed now, but I'll look it over tomorrow and make a new release for you.

Issue #2104, opened 2023-07-22
Add macOS as operating system


postlund:
### What feature would you like?

Since macOS can act as AirPlay receiver, it makes sense to add that as operating system as well.

### Describe the solution you'd like

I don't think there are any better way to detect this than looking at `model` (e.g. `MacBookAir10,1` maps to macOS). Just need to figure out all valid models. So far I know about these:

* MacBookAirX,Y
* iMacX,Y
* MacminiX,Y
* MacBookProX,Y
* MacX,Y
* MacProX,Y

Not sure if all of them supports macOS 12 or later (where AirPlay receiver is built in), but there's no harm adding all of them since they all map to macOS.

Sources:
https://www.theiphonewiki.com/wiki/Models#Mac
https://mrmacintosh.com/list-of-mac-boardid-deviceid-model-identifiers-machine-models/

### Any other information to share?

No

postlund:
Fixed by #2113

PR #2113, opened 2023-07-28
conf: Add macOs as operating system

postlund:
Detection has been added fir AirPlay and RAOP.

Relates to #2104


stibbzy:
@postlund Hi there. I’ve finally been able to add my MacBook (using the AirPlay Receiver feature) to the ATV integration in Home Assistant. However, the two entities it produces show no actual activity. The media player entity’s status is constantly ‘idle’ no matter whether it is receiving an Airplay stream or not. 

Is this expected operation, and just a shortcoming of the protocol? It just seems that as it is right now, the entities don’t have any utility at all. 

Many thanks!

postlund:
@stibbzy It's expected behavior unfortunately. As far as I have managed to figure out, there's nothing in AirPlay that allows you to get metadata from an AirPlay receiver that already plays something. It is possible to see the play state via Zeroconf properties (not implemented in pyatv though). Some people seem to suggest that they see this, but I'm not sure how it is possible. So current state is basically what I can do in pyatv.

stibbzy:
@postlund Ahh I see, thanks! So am I right in assuming that whatever protocol (assuming it's MRP) that allows us to see Apple TV and HomePod metadata isn't present in the MacOS implementation of Airplay Receiver? (i.e. MRP is tvOS only?)

PR #2122, opened 2023-08-03
Fallback to sending unicast PTR queries when multicast is broken or packets are being dropped

bdraco:
fixes https://github.com/home-assistant/core/issues/95768

I came up with a pretty good way to test this.  I moved my sonos roam to the edge of wifi range to the point were it had significant packet loss and removed the filters in HA. I'm now able to add the roam when the wifi has significant packet loss. This is effectively the same problem as why Passive Observation Of Failures (POOF) is unviable for setups where the responders are on Wi-Fi

By design we only send one unicast packet to do the fallback query to minimize the network and processing impact.  If the network is flakey Home Assistant is going to retry setup again anyways and the next attempt is likely to be successful. 

This is a bit complex because the unicast scanner takes a list of hosts so we have to send to handle that case even though in practice we only ever send to one. We may need this behavior in the future so I didn't remove it and make it a single host only.

### DIFF
--- a/pyatv/core/scan.py
+++ b/pyatv/core/scan.py
@@ -563,7 +563,6 @@ class ZeroconfUnicastScanner(ZeroconfScanner):
                 if infos_by_type := infos_by_address_type.get(address):
                     infos_by_type[info.type] = info
                     self._set_or_get_address_to_device_name(address, info)
-                    break
 
         # Device info is special because it does not have an address
         device_name_to_address = self.device_name_to_address
--- a/pyatv/core/scan.py
+++ b/pyatv/core/scan.py
@@ -516,7 +516,11 @@ class ZeroconfUnicastScanner(ZeroconfScanner):
     def _discard_completed_needed_types(
         self, infos: List[AsyncServiceInfo], loaded_from_cache: bool
     ) -> List[AsyncServiceInfo]:
-        """Discard needed type that we have info for."""
+        """Discard needed type that we have info for.
+        
+        Returns a list of incomplete infos that we still need to
+        send queries for.
+        """
         zeroconf = self.zeroconf
         incomplete_infos: List[AsyncServiceInfo] = []
         needed_types_by_address = self.needed_types_by_address
--- a/pyatv/core/scan.py
+++ b/pyatv/core/scan.py
@@ -526,7 +526,7 @@ class ZeroconfUnicastScanner(ZeroconfScanner):
         ] = {host: {f"{type_}.": None for type_ in self._services} for host in hosts}
 
     def _send_ptr_queries(
-        self, address: IPv4Address, needed_types: Set[str]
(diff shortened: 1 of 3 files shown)


pyatv/core/scan.py
```diff
@@ -480,3 +456,203 @@ async def process(self, timeout: int) -> None:
...
+        out = DNSOutgoing(_FLAGS_QR_QUERY)
+        # Note that the multicast flag here is True even though we are sending
+        # unicast queries. We still are speaking mdns and will be happy to get
+        # a multicast response back.
+        for type_ in needed_types:
+            question = DNSQuestion(type_, _TYPE_PTR, _CLASS_IN)
+            question.unicast = True
+            out.add_question(question)
+        self.zeroconf.async_send(out, str(address))
+
```
bdraco, line 541: This is a bit of a big hammer but its unicast to a single host so only HA has to pay to process it and we aren't flooding the network with requests so it seems like an ok trade off

bdraco:
profile

bdraco:
performance fix https://github.com/python-zeroconf/python-zeroconf/pull/1211

bdraco:
hass bump https://github.com/home-assistant/core/pull/97745

bdraco:
with the performance changes in zc
zc_changes

bdraco:
performance trade off to solve this seems ok now

bdraco:
@postlund I ended up going down the rabbit hole quite a bit on this one.  This should fix many of the users who have flakey multicast or devices that drop udp packets due to wifi interference/range issues.

postlund:
@bdraco Yeah, I noticed 😉 it's great to see improvements in this area though, seems to pop up issues every now and then regarding this. So 👍 I will try to look over this tonight. Do you want to wait for confirmation in https://github.com/home-assistant/core/issues/95768 before I make a release?

bdraco:
> @bdraco Yeah, I noticed 😉 it's great to see improvements in this area though, seems to pop up issues every now and then regarding this. So 👍 I will try to look over this tonight. Do you want to wait for confirmation in https://github.com/home-assistant/core/issues/95768 before I make a release?

I ran the Sonos roam at the edge of Wi-Fi range in a reload loop overnight and it stays stable now.  It also improved the situation one of my (Wi-Fi connected) HomePod that needs to try twice every so often.

If we don't solve 95768 with this, we likely need more information before I do another run at it.

Since it will at least solve the Sonos roam case it's good to release.

Cheers

bdraco:
Everything has been rock solid now on all my production setups with the HomePods and Sonos devices moved to the edge of Wi-Fi range

It would be great to get a new release

bdraco:
Also noticed a nice traffic reduction since it's not having to retry anymore. So even with the unicast ptr probe the net result is less traffic

postlund:
Awesome! Will push for it!

bdraco:
thanks

PR #2236, opened 2023-10-09
add py.typed to mark pyatv as PEP-561 compliant

stackia:
Hi! I'm working on enabling [strict typing check](https://github.com/home-assistant/core/blob/dev/.strict-typing) for home assistant `apple_tv` integration. Currently, pyatv is not declared as PEP-561 compliant which causes mypy to [ignore all imports from pyatv](https://github.com/home-assistant/core/actions/runs/6454672708/job/17521010925?pr=101688).

Since most of pyatv are already typed, we can safely add a `py.typed` to mark this library as PEP-561 compliant.

See: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

### DIFF
--- a/pyatv/interface.py
+++ b/pyatv/interface.py
@@ -510,7 +510,7 @@ class Playing(ABC):
         self._content_identifier = content_identifier
         self._post_process()
 
-    def _post_process(self):
+    def _post_process(self) -> None:
         if self._position:
             # Make sure position never is negative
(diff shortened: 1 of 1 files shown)


postlund:
Didn't know about this, great addition. Thanks! I'm gonna try to catch up with things and have a new release out later today (best case) or tomorrow (more likely case).

PR #2241, opened 2023-10-13
Filter out invalid addresses from zeroconf

bdraco:
https://github.com/home-assistant/core/issues/101948


pyatv/core/scan.py
```diff
@@ -328,6 +329,29 @@ def _name_without_type(name: str, type_: str) -> str:
...
+) -> List[IPv4Address]:
+    """Return valid IPv4 addresses from service info.
+
+    Older airport devices (e.g. Airport Express) can return
+    link-local addresses in the service info after the valid
+    addresses. This function will filter out those addresses.
+    """
+    addresses = info.ip_addresses_by_version(IPVersion.V4Only)
+    if TYPE_CHECKING:
+        ipv4_addresses = cast(List[IPv4Address], addresses)
```
bdraco, line 343: Guarded the type checking here since its creating a new `List[]` typing object each time

Issue #2261, opened 2023-10-23
Add backwards compatibility with pydantic v1


postlund:
### What to change?

Seems it's not possible to use pydantic v2 in Home Assistant at the moment. Currently only the extra types package seems to be the problem, so I need to get rid of that and add some shims. More info here:

https://github.com/home-assistant/core/issues/99218
https://github.com/AngellusMortis/pyunifiprotect/pull/297

postlund:
Fixed in 0.14.3.

PR #2262, opened 2023-10-23
Use pydantic v1 compatibility when with pydantic v2

postlund:
As described in #2261.

### DIFF
--- a/base_versions.txt
+++ b/base_versions.txt
@@ -6,7 +6,7 @@ ifaddr==0.1.7
 mediafile==0.8.1
 miniaudio==1.45
 protobuf==4.23.4
-pydantic==2.4.0
+pydantic==1.10.10
 requests==2.23.0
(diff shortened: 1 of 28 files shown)


base_versions.txt
```diff
@@ -6,8 +6,7 @@ ifaddr==0.1.7
 mediafile==0.8.1
 miniaudio==1.45
 protobuf==4.23.4
-pydantic-extra-types==2.1.0
-pydantic-settings==2.0.2
+pydantic==1.10.13
```
joostlek, line 9: Pydantic is currently pinned to 1.10.12 in HA. Not sure how the whole dependency of dependencies song and dance works and if this will be a problem, but just FYI

postlund: Good catch, I will update that. The versions in this file be translated into ">=", so at least the versions specified here are required.

Issue #2268, opened 2023-10-25
Python 3.11 / 3.12 atvscript push_updates throwing errors


maxileith:
### Describe the bug

Hello everybody,

I want to utilize the `atvscript push_updates` for a project of mine. I have noticed, that `atvscript push_updates` does not work when using Python 3.11 or 3.12. With Python 3.10 everything works just fine. I have attached the logs below. `atvremote push_updates` does run without any errors.

### Error log

```log
❯ atvscript -s 42.42.42.42 push_updates --companion-credentials xxxx --airplay-credentials xxxx
{"result": "success", "datetime": "2023-10-25T16:46:33.635718+02:00", "power_state": "on"}
{"result": "success", "datetime": "2023-10-25T16:46:33.635767+02:00", "output_devices": [{"name": "Wohnzimmer (2)", "identifier": "82091D04-8D8E-4CF8-8FAA-4B7FC9979EB9"}]}
{"result": "success", "datetime": "2023-10-25T16:46:33.636150+02:00", "hash": "ca496c14642c78af6dd4250191fe175f6dafd72b4c33bcbab43c454aae051da1", "media_type": "unknown", "device_state": "idle", "title": null, "artist": null, "album": null, "genre": null, "total_time": null, "position": null, "shuffle": "off", "repeat": "off", "series_name": null, "season_number": null, "episode_number": null, "content_identifier": null, "app": null, "app_id": null}
[...]

PR #2270, opened 2023-10-26
Wrap coroutines passed to asyncio.wait with create_tasks because asyncio.wait does not support passing coroutines since python 3.11

maxileith:
Hello everybody,

yesterday I reported the bug #2268.

Today, I have taken a look at the problem. It was a fairly easy fix. The problem was that [asyncio.wait](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait) forbidds to pass coroutine objects directly since Python 3.11.

To fix this issue, the coroutines can just be wrapped in a task by utilizing [asyncio.create_task](https://docs.python.org/3/library/asyncio-task.html#asyncio.create_task).

If there are any questions, feel free to ask :).


maxileith:
@postlund I have updated the PR. Could you please trigger the automated tests again?

postlund:
Looks reasonable to me, I'll try to make a release as soon as I can. Thanks! 👍

PR #2281, opened 2023-11-08
Changing the featured Homebridge plugin to a more recent and maintained one.

maxileith:
Hi @postlund ,

I am currently developing a homebridge plugin that is already in a very solid state. I would suggest to replace the currently featured but not maintained homebridge plugin (seems like it when looking at the open issues) from @cristian5th with this maintained one. Also my plugin is more native, since it does not provide a configuration for another homebridge plugin but is a homebridge plugin by itself.

@cristian5th you are welcome to join the discussion


postlund:
This makes sense to me, thanks for updating and your work on the updated homebridge plugin! 👍

PR #2282, opened 2023-11-08
Output all identifiers and mac address when scanning with atvscript

maxileith:
Hi @postlund ,

for the use-case I am currently working on, I use Airplay MAC-Adresses as an identifier https://github.com/maxileith/homebridge-appletv-enhanced/issues/20. Therefore, I always need to get the MAC-Address as an identifier from `atvscript`. However, currently the returned identifier from `atvscript scan` is not always the MAC-Address. Therefore, I would suggest to add an attribute `all_identifieres` to the output json. By not touching `identifier` of the output json, backward compatability is given.

I'd love to hear your thoughts on that.

### DIFF
--- a/docs/documentation/atvscript.md
+++ b/docs/documentation/atvscript.md
@@ -73,7 +73,12 @@ $ atvscript scan
       "name": "Vardagsrum",
       "address": "10.0.10.81",
       "identifier": "xxx",
-      {
+      "all_identifiers": [
+        "xxx",
+        "xxx",
+        "xxx"
+      ],
+      "device_info": {
         "model": "Gen4K",
         "model_str": "Apple TV 4K",
         "operating_system": "TvOS",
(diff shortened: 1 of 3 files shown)


maxileith:
I have updated the pull request to include the MAC-Address in the device infos since I noticed that this is available on the atv object anyways

maxileith:
Hi everybody,

fyi, I am still waiting for the PR to get reviewed. I haven't forgotten the bug. Seems like the owner of pyatv is on vacation since there are no commits on this repository since a month or so ...

maxileith:
Hi @postlund ,

Can you do a ruff estimation on when you will have the time to look at the PR? I don't want to stress, I am just curious and would like to evaluate wether I will wait for the PR or do a workaround 😬

postlund:
@maxileith Sorry about the delay, I will try to get around doing that this week!

postlund:
Sorry for not getting back sooner on this, but it looks good to me! 👍 I'll see if there's anything else to include, otherwise I can make a minor release soon with this change.

maxileith:
@postlund don't worry about the late response. We all have private lifes besides our passion for FOSS 😉

PR #2338, opened 2024-01-26
Simplify ChaCha20 Poly1305 implementation

bdraco:
similar change to
https://github.com/esphome/aioesphomeapi/blob/c9227edf7fcee0f175c9f31c024ba71b0a0fac26/aioesphomeapi/_frame_helper/noise.py#L31
https://github.com/ikalchev/HAP-python/blob/dev/pyhap/hap_crypto.py#L17

`pack` is a bit faster than `to_bytes`

### DIFF
--- a/pyatv/support/chacha20.py
+++ b/pyatv/support/chacha20.py
@@ -1,9 +1,9 @@
 """Transparent encryption layer using Chacha20_Poly1305."""
+from functools import partial
+from struct import Struct
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


bdraco:
Thanks for the 👀

postlund:
This unfortunately breaks Companion that uses 12 bytes as nonce length 😳

bdraco:
No worries, guess we need to revert. Sorry I missed that.

bdraco:
I guess we could subclass it instead and have a separate fast pack for that

bdraco:
https://github.com/postlund/pyatv/pull/2412

PR #2362, opened 2024-02-26
add fast forward and rewind support for companion protocol

nfreeze:


pyatv/const.py
```diff
@@ -5,7 +5,7 @@
 
 MAJOR_VERSION = "0"
 MINOR_VERSION = "14"
-PATCH_VERSION = "5"
+PATCH_VERSION = "4"
```
postlund, line 8: Please leave out this change.

PR #2363, opened 2024-02-26
add fast forward and rewind support for companion protocol

nfreeze:


nfreeze:
I closed the previous PR and made the requested change and a couple of others

postlund:
> I closed the previous PR and made the requested change and a couple of others

You can just `push --force` to the branch and it will update in the PR (as a PR is just a branch), no need for a new PR.

If you check the functional tests for companion, I'd like you to add the new feature constants so they are tested as well.

https://github.com/postlund/pyatv/blob/4b00261b4c08ba1e13db605b109eb1a0df700ffe/tests/protocols/companion/test_companion_functional.py#L46

I'll try to make a full review within the next few days, when I'm in front of a computer. Thanks for the work so far 👌

nfreeze:
Ah, thanks for the tip. I'm new to the PR workflow so figured I was doing
it wrong :)  Will update the companion test in a bit.

nfreeze:
I am closing this because I didn't realize you can do right=2 and left=2 in atvremote to simulate fast forward and rewind

postlund:
That might be true for Apple TVs ringing tvOS, yes. But it won't work for older models nor HomePods or other AirPlay receivers. So I still see a value in adding this.

postlund:
> Thanks for adding this feature! Let me know if I can help out somehow.
> 
> > you can do right=2 and left=2 in atvremote to simulate fast forward and rewind
> 
> Could you please elaborate? If I use `remote_control.right()` / `remote_control.left()` on an Apple TV 4K (gen 2), tvOS 17.3, it only skips forward and backward. Or do you mean something else?

https://pyatv.dev/development/control/#input-actions

nfreeze:
If you single press right it skips forward 10 seconds but if you press and
hold right (right=2 in atvremote) it goes into fast forward.

PR #2397, opened 2024-05-10
Check companion pairing authentication status code

ninjaoxygen:
Adds the status code check for companion pairing authentication, raises an AuthenticationError if they are not the expected status code.

### DIFF
--- a/pyatv/protocols/companion/auth.py
+++ b/pyatv/protocols/companion/auth.py
@@ -158,7 +158,10 @@ class CompanionPairVerifyProcedure(PairVerifyProcedure):
             },
         )
 
-        # TODO: check status code
+        # check status code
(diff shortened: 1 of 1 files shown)


pyatv/protocols/companion/auth.py
```diff
@@ -158,7 +158,10 @@ async def verify_credentials(self) -> bool:
             },
         )
 
-        # TODO: check status code
+        # check status code
+        if resp['_pd'][2] != 4:
```
postlund, line 162: I'm a bit uncertain what this check really is? Isn't `2` here the same as https://github.com/postlund/pyatv/blob/8d78f516afe0e738d5eedc054a22fc27b6900d0c/pyatv/auth/hap_tlv8.py#L18 ?

PR #2402, opened 2024-06-18
Added touch gesture to companion protocol

albaintor:
Hi, I have added the implementation of touch gestures and updated the documentation.

I have tested it successfully.
To test it :
```python
atv = await pyatv.connect(config, loop)
companion = cast(FacadeRemoteControl, atv.remote_control).get(Protocol.Companion)
if companion:
    api: CompanionAPI = companion.api
    # Simulate a gesture from x=100, y=100 to x=1000, y=1000 in 1000 milliseconds
    await api.touch_gesture(100, 100, 1000, 1000, 1000)
```


IsakTheHacker (approved):

albaintor:
Hi, I have added the new interface and its implementation including the facadeAppleTV class

IsakTheHacker:
Hi, amazing work you've done here! Have been waiting for this for so long. Do you think it would be possible to have this working for real time gestures, just like the way it works for the Siri Remote and in control center on iOS? I have no idea how such an interface would work or even look like though, but that would be very awesome. Everything looks exceptionally good by the way and I hope @postlund can take a look at this soon

albaintor:
https://github.com/postlund/pyatv/assets/118518828/a3bd72ce-8726-42b6-a017-05d64fac83fa

I have made a tool to simulate the touchpad gestures and the right algorithm behind

pyatv/core/facade.py
```diff
@@ -600,6 +600,31 @@ def playstatus_error(self, updater, exception: Exception) -> None:
...
+
+    @shield.guard
+    async def touch_gesture(self, start_x: int, start_y: int, end_x: int, end_y: int, duration_ms: int) -> None:
+        """Generate a touch gesture from start to end x,y coordinates"""
+        return await self.relay("touch_gesture")(start_x=start_x, start_y=start_y,
+                                                 end_x=end_x, end_y=end_y, duration_ms=duration_ms)
+
+    @shield.guard
+    async def touch_event(self, x: int, y: int, mode: int) -> None:
+        """Generate a touch gesture from start to end x,y coordinates"""
```
postlund, line 619: Forgot to update this one

pyatv/core/facade.py
```diff
@@ -832,6 +858,12 @@ def keyboard(self) -> interface.Keyboard:
         """Return keyboard interface."""
         return cast(interface.Keyboard, self._interfaces[interface.Keyboard])
 
+    @property  # type: ignore
+    @shield.guard
+    def touchgestures(self) -> interface.TouchGestures:
```
postlund, line 863: I think we can go with just `touch` here instead.

pyatv/interface.py
```diff
@@ -1249,6 +1249,34 @@ async def text_set(self, text: str) -> None:
...
+        :param start_x: Start x coordinate
+        :param start_y: Start y coordinate
+        :param end_x: End x coordinate
+        :param end_y: Endi x coordinate
+        :param duration_ms: Time in milliseconds to reach the end coordinates
+        """
+        raise exceptions.NotSupportedError()
+
+    @feature(64, "TouchEvent", "Touch event to given coordinates.")
+    async def touch_event(self, x: int, y: int, mode: int) -> None:
```
postlund, line 1266: Please create an enum for `mode` to maintain proper typing.

pyatv/protocols/companion/__init__.py
```diff
@@ -461,6 +464,39 @@ async def text_set(self, text: str) -> None:
         await self.api.text_input_command(text, clear_previous_input=True)
 
 
+class CompanionTouchGestures(TouchGestures):
+    """Implementation of touch gesture API."""
+
+    def __init__(self, api: CompanionAPI) -> None:
+        """Initialize a new CompanionRemoteControl."""
```
postlund, line 471: `ComapnionRemoteControl` -> `CompanionTouchGeatures`

pyatv/protocols/companion/api.py
```diff
@@ -350,3 +417,15 @@ async def fetch_attention_state(self) -> SystemStatus:
             raise exceptions.ProtocolError("missing content")
 
         return SystemStatus(content["state"])
+
+    async def _touch_start(self) -> Mapping[str, Any]:
+        """ Subscribe to touch gestures """
+        self._base_timestamp = time.time_ns()
+        response = await self._send_command("_touchStart",
+            {"_height": TOUCHPAD_HEIGHT, "_tFl": 0, "_width": TOUCHPAD_WIDTH})
+        await asyncio.gather(*self.dispatch("_touchStart", response.get("_c", {})))
```
postlund, line 426: I don't see why you need to use `asyncio.gather` here?

pyatv/protocols/companion/protocol.py
```diff
@@ -184,6 +184,7 @@ def send_opack(self, frame_type: FrameType, data: Dict[str, Any]) -> None:
             self._xid += 1
 
         _LOGGER.debug("Send OPACK: %s", data)
+        print(data)
```
postlund, line 187: Please remove this

pyatv/scripts/atvui.py
```diff
@@ -0,0 +1,128 @@
+import argparse
```
postlund, line 1: I think we should deal with the UI in a separate PR. I'm all for including it since it does not add any additional dependencies though. Will  have to review and test it separately.

postlund (changes requested):
A few comment but it looks very good, nice addition! 👍

postlund:
A section in the `atvremote` documentation with some examples would be great as well!

pyatv/core/facade.py
albaintor: Ok done

pyatv/core/facade.py
albaintor: Ok done

pyatv/interface.py
albaintor: done

pyatv/protocols/companion/__init__.py
albaintor: done

pyatv/protocols/companion/api.py
albaintor: correct not needed

pyatv/protocols/companion/protocol.py
albaintor: Done

pyatv/scripts/atvui.py
albaintor: Ok I removed the script

postlund:
One last thing that I forgot about... Some kind of basic tests for the added API methods are needed. Only a "good case" is needed, just to ensure we verify something. Should be fairly easy to add in https://github.com/postlund/pyatv/blob/master/tests/protocols/companion/test_companion_functional.py

albaintor:
> A section in the `atvremote` documentation with some examples would be great as well!

Yes I will add that

albaintor:
> One last thing that I forgot about... Some kind of basic tests for the added API methods are needed. Only a "good case" is needed, just to ensure we verify something. Should be fairly easy to add in https://github.com/postlund/pyatv/blob/master/tests/protocols/companion/test_companion_functional.py

I don't think this is possible : touch events are sent with no feedback by Appletv

postlund:
> > One last thing that I forgot about... Some kind of basic tests for the added API methods are needed. Only a "good case" is needed, just to ensure we verify something. Should be fairly easy to add in https://github.com/postlund/pyatv/blob/master/tests/protocols/companion/test_companion_functional.py
> 
> I don't think this is possible : touch events are sent with no feedback by Appletv

I have implanted fake "server sides" of each protocol. So you can just store the events and verify what you received. No interaction with an Apple TV is used (it's just unit tests).

albaintor:
> > > One last thing that I forgot about... Some kind of basic tests for the added API methods are needed. Only a "good case" is needed, just to ensure we verify something. Should be fairly easy to add in https://github.com/postlund/pyatv/blob/master/tests/protocols/companion/test_companion_functional.py
> > 
> > 
> > I don't think this is possible : touch events are sent with no feedback by Appletv
> 
> I have implanted fake "server sides" of each protocol. So you can just store the events and verify what you received. No interaction with an Apple TV is used (it's just unit tests).

Ok got it, I am on it

albaintor:
Hi,
all done. I'll let you review. I have also updated the atvremote script to support touch calls

albaintor:
As an example :
```shell
# Horizontal touch gesture from left to right in 200ms
atvremote.py --id <appletvid> touch=0,0,800,0,200
# Touch click
atvremote.py --id <appletvid> touch_click
```

postlund:
I'll try to look at it tonight, great work so far! Also remember that we should add some developer documentation as well at some point, in case someone wants to use it (a subpage here https://pyatv.dev/development/).

albaintor:
> I'll try to look at it tonight, great work so far! Also remember that we should add some developer documentation as well at some point, in case someone wants to use it (a subpage here https://pyatv.dev/development/).

Yes I a little new to python's coding (since this year), however I found your code very well organized apart all the offered features for all-in-one library.
I got a little struggled though by the fact that I had to apply modifications in several files (don't remember exactly but : companion/api, companion/protocol, interfaces, and facadeappletv)

pyatv/protocols/companion/api.py
```diff
@@ -282,6 +290,59 @@ async def hid_command(self, down: bool, command: HidCommand) -> None:
...
+        :param end_y: Endi x coordinate
+        :param duration_ms: Time in milliseconds to reach the end coordinates
+        """
+        end_time = time.time_ns() + duration_ms * 1000000
+        x: float = float(start_x)
+        y: float = float(start_y)
+        await self.hid_event(int(x), int(y), HidEventMode.Press)
+        sleep_time = float(TOUCHPAD_DELAY_MS / 1000)
+        current_time = time.time_ns()
+        while current_time < end_time:
```
thiccaxe, line 319: nitpick, but I wonder if using a simple for loop would make more sense here:
Calculate the total amount of iterations needed by dividing the duration_ms the delay or something of that nature.

thiccaxe:
I'd also ask you to take a look at "press and hold"
This is the same as regular click, except:
press:
```js
{'_i': '_hidC', '_x': 4067597582, '_btHP': False, '_c': {'_hBtS': 1, '_hidC': 6}, '_t': 2}
```
wait, say 1 second or something, then:
```js
{'_i': '_hidT', '_x': 4067597583, '_c': {'_ns': 86787906162458, '_tFg': 1, '_cx': 1000, '_tPh': 5, '_cy': 1000}, '_t': 1}
{'_i': '_hidC', '_x': 4067597584, '_btHP': False, '_c': {'_hBtS': 2, '_hidC': 6}, '_t': 2}
```

pyatv/protocols/companion/api.py
albaintor: No because you cannot be sure how long it will take to send the events (depending on client device activity and cpu skills) and the timestamp sent has to be very precise (in nanoseconds). But maybe this is not a big deal not to be very precise with this... in that way we are sure to finish after the exact given duration and coordinates

albaintor:
> I'd also ask you to take a look at "press and hold" This is the same as regular click, except: press:
> 
> ```js
> {'_i': '_hidC', '_x': 4067597582, '_btHP': False, '_c': {'_hBtS': 1, '_hidC': 6}, '_t': 2}
> ```
> 
> wait, say 1 second or something, then:
> 
> ```js
> {'_i': '_hidT', '_x': 4067597583, '_c': {'_ns': 86787906162458, '_tFg': 1, '_cx': 1000, '_tPh': 5, '_cy': 1000}, '_t': 1}
> {'_i': '_hidC', '_x': 4067597584, '_btHP': False, '_c': {'_hBtS': 2, '_hidC': 6}, '_t': 2}
> ```

I was wondering if I had covered all the touch gestures and I missed it... to be done indeed

pyatv/protocols/companion/api.py
thiccaxe: Well, from what I can guess at,
the _ns flag records the actual timestamp when the touch event was recorded on the client. This is to mitigate network lag. The Apple TV looks at the _ns field to determine when the touch actually occurred (relative the previous touches). So _ns can be precalculated and doesn't depend on the actual time when the packet is sent.

albaintor: I am not so sure about that : we are talking about 10ms which is very low and may be affected by cpu charge and affect the movement. 
You have to consider that this library may be used on low end device. 
I am not opposed to simplification but this has to be tested and confirmed

albaintor:
> I'll try to look at it tonight, great work so far! Also remember that we should add some developer documentation as well at some point, in case someone wants to use it (a subpage here https://pyatv.dev/development/).

Hi,
I have fixed a bug in touch click (delay needed between commands otherwise it won't work all the time) and added the missing touch_hold command. Please let me know if there is something  more to be done.

Also Pierre I am interested in adding voice commands to the library (send siri commands in audio or text format). Your proxy program could sniff those requests too ? Thanks

postlund:
Great, I will take a look! Trying to catch up without work ask I can make release with this soon too.

I have not tested it, but I expect that it should work. I guess you will have to test. One thing to keep in min though is that we should try to re-use as much work as possible from AirPlay streaming when providing audio/voice. I expect that the user should be able to pick synthesizer and just provide a text string to speak or just provide a pre-recorded file. We need to be flexible here.

albaintor:
Hi Pierre,

There is a regression on your recent changes : the modifications applied on support/chacha20.py broke the support of companion protocol
https://github.com/postlund/pyatv/commit/d5a29d7efbbaf05920b08a4e54bea178ddf883b2

If I revert those changes it works again

Thanks

De : Pierre Ståhl ***@***.***>
Envoyé : samedi 22 juin 2024 10:46
À : postlund/pyatv ***@***.***>
Cc : albaintor ***@***.***>; Author ***@***.***>
Objet : Re: [postlund/pyatv] Added touch gesture to companion protocol (PR #2402)


Great, I will take a look! Trying to catch up without work ask I can make release with this soon too.

I have not tested it, but I expect that it should work. I guess you will have to test. One thing to keep in min though is that we should try to re-use as much work as possible from AirPlay streaming when providing audio/voice. I expect that the user should be able to pick synthesizer and just provide a text string to speak or just provide a pre-recorded file. We need to be flexible here.

-

postlund:
I see the regression as well and I know why, will fix it later.

You can trigger Siri by holding the power button with the remote up (like you do to trigger Siri in iOS) I think.

thiccaxe:
Voice is easy, the hard part is figuring out the format apple uses. The best option for now might be to capture packets from an I-device and then when requested re-send those with pyatv. Atleast as an initial implementation

thiccaxe:
https://github.com/postlund/pyatv/issues/2325#issuecomment-2001059905

thiccaxe:
AFAIK you need Siri enabled on the Apple TV AND iPhone for it to work...

postlund:
I'm pretty sure I did dumps of some voice interaction back in the days over MRP. Must have misplaced the protobufs I made for that, as I'm pretty sure there were mentions of things such as format (e.g. PCM8/16/24) and actual audio samples. Would expect something similar would be used here as well.

pyatv/const.py
```diff
@@ -434,3 +434,23 @@ class FeatureName(Enum):
...
+    """Touch event to given coordinates."""
+
+    TouchClick = 65
+    """Touch click command."""
+
+    TouchHold = 66
+    """Hold touch command."""
+
+
+class HidEventMode(Enum):
```
postlund, line 451: I think we sold call this something like `TouchAction` as `HID` is just an implementation detail in this context. Could change to something else in the future.

postlund: Do we know what 2 is btw?

pyatv/core/facade.py
```diff
@@ -601,6 +601,36 @@ def playstatus_error(self, updater, exception: Exception) -> None:
...
+        interface.TouchGestures.__init__(self)
+
+    @shield.guard
+    async def touch(self, start_x: int, start_y: int, end_x: int, end_y: int, duration_ms: int) -> None:
+        """Generate a touch gesture from start to end x,y coordinates"""
+        return await self.relay("touch")(start_x=start_x, start_y=start_y,
+                                         end_x=end_x, end_y=end_y, duration_ms=duration_ms)
+
+    @shield.guard
+    async def touch_event(self, x: int, y: int, mode: HidEventMode) -> None:
```
postlund, line 619: Maybe go with `touch_action` here? I would read "event" more like something that happened on the device rather than something that *I* did.

tests/fake_device/companion.py
```diff
@@ -57,6 +57,14 @@
 }
 
 
+class HidEvent:
```
postlund, line 60: You can replace this with a dataclass.

pyatv/core/facade.py
```diff
@@ -601,6 +601,36 @@ def playstatus_error(self, updater, exception: Exception) -> None:
...
+class FacadeTouchGestures(Relayer, interface.TouchGestures):
+    """Facade implementation for touch gestures handling."""
+
+    def __init__(self, core_dispatcher: CoreStateDispatcher):
+        """Initialize a new FacadeTouchGestures instance."""
+        Relayer.__init__(self, interface.TouchGestures, DEFAULT_PRIORITIES)
+        interface.TouchGestures.__init__(self)
+
+    @shield.guard
+    async def touch(self, start_x: int, start_y: int, end_x: int, end_y: int, duration_ms: int) -> None:
```
postlund, line 613: This is more of a swipe I think?

postlund (changes requested):

pyatv/const.py
thiccaxe: I have not seen it as of now. However, I believe that "Hold" ought to be renamed to "Drag" to not confuse it with "press and *hold*"

pyatv/scripts/atvremote.py
```diff
@@ -833,6 +835,8 @@ def _parse_args(cmd, args):
             return [InputAction(args[0])]
         if cmd == "set_volume":
             return [float(args[0])]
+        if cmd == "touch":
```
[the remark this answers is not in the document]
albaintor: you're right, I don't understand why it didn't work at start but probably because I pushed args wrongly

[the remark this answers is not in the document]
albaintor: I removed these lines

pyatv/core/facade.py
```diff
@@ -601,6 +601,36 @@ def playstatus_error(self, updater, exception: Exception) -> None:
...
+        return await self.relay("touch")(start_x=start_x, start_y=start_y,
+                                         end_x=end_x, end_y=end_y, duration_ms=duration_ms)
+
+    @shield.guard
+    async def touch_event(self, x: int, y: int, mode: HidEventMode) -> None:
+        """Generate a touch event to end x,y coordinates"""
+        return await self.relay("touch_event")(x=x, y=y, mode=mode)
+
+    @shield.guard
+    async def touch_click(self):
```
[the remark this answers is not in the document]
albaintor: Ok I merged the 2 methods hold & click in one with action parameter

pyatv/core/facade.py
albaintor: Ok renamed

pyatv/const.py
albaintor: I didn't see any message with 2, so only guess at this state

pyatv/core/facade.py
albaintor: I renamed the method to touch_swipe

tests/fake_device/companion.py
albaintor: yes I know now what a dataclass is thanks :-)

albaintor:
I have committed changes, you can review them.

albaintor:
> [#2325 (comment)](https://github.com/postlund/pyatv/issues/2325#issuecomment-2001059905)

Hi, I have managed to make siri commands work using text but only if I enable "write to siri" in accessibility options : this raises a text field that I can play with. Otherwise it expects a voice command. I didn't find out how to provide text commands with another method (is it possible ?)
Here is the sample code in the companion API : for example if you call _test_siri("time") it will display current time on AppleTV
Otherwise your link is not enough to digg into voice commands : I know how to raise a siri start/stop command, but the tough part is the requests with payloads (audio protocol...)


```python
    async def _test_siri(self, command):
        await self._siri_start()
        await asyncio.sleep(1)
        await self._siri_stop()
        await self.text_input_command("", clear_previous_input=True)
        await asyncio.sleep(0.1)
        await self.text_input_command(command, clear_previous_input=True)
        await self.hid_command(True, HidCommand.Down)
        await self.hid_command(False, HidCommand.Down)
        await asyncio.sleep(0.1)
        await self.hid_command(True, HidCommand.Down)
        await self.hid_command(False, HidCommand.Down)
        await asyncio.sleep(0.1)
        await self.hid_command(True, HidCommand.Down)
        await self.hid_command(False, HidCommand.Down)
        await asyncio.sleep(0.1)
        await self.touch_click(InputAction.SingleTap)
        await asyncio.sleep(0.1)

    async def _siri_start(self) -> Mapping[str, Any]:
        response = await self._send_command("_siriStart", {})
        await asyncio.gather(*self.dispatch("_siriStart", response.get("_c", {})))
        return response

    async def _siri_stop(self) -> None:
        await self._send_command("_siriStop", {})

**Comment by thiccaxe** [comment]
You should be able to submit using the text commands.

**Comment by albaintor** [comment]
> You should be able to submit using the text commands.

I am not sure : maybe there is a command to submit input text (instead of moving cursor to submit button and click). However is there a dedicated command to send the submitted text as a siri command (when accessibility setting is disabled) ? Not so sure

**Review by IsakTheHacker — CHANGES_REQUESTED** [review]

**Comment by albaintor** [comment]
Hi @postlund will you have time to review the last updates ?
Thank you

File: pyatv/const.py (hunk 9 above)
Comment (albaintor) [reply pyatv/const.py]: Renamed

**Comment by postlund** [comment]
Sorry for the delay, I will review it later tonight before going to bed.

File: pyatv/interface.py [hunk 15]
```diff
@@ -1249,6 +1250,45 @@ async def text_set(self, text: str) -> None:
         raise exceptions.NotSupportedError()
 
 
+class TouchGestures(ABC):
+    """Base class for touch gestures."""
+
+    @feature(63, "TouchSwipe", "Touch swipe from given coordinates and duration.")
+    async def touch_swipe(
```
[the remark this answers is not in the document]
postlund: I tend to agree with @IsakTheHacker here. There has been some mixing in this regards in the past, but I would prefer to keep simple names when possible. So changing to just `swipe` is a good idea.

pyatv/interface.py
```diff
@@ -1249,6 +1250,45 @@ async def text_set(self, text: str) -> None:
...
+        self, start_x: int, start_y: int, end_x: int, end_y: int, duration_ms: int
```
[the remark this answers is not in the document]
postlund: I believe this is fine for simpler methods like this. My suggestion would be to add another method, accepting a sequence of something like "swipe"-objects (basically a dataclass with all arguments to this method as members). That would allow for definition of swipe patterns, something like this:

```python
SWIPE_RIGHT_LEFT = [
    Swipe(0, 0, 500, 0, 100),  # X: 0 -> 500 (right)
    Swipe(500, 0, 0, 0, 100),  # X: 500 -> 0 (left)
]

atv.touch.swipe_sequence(SWIPE_RIGHT_LEFT)
```

I'm up for discussion here regarding names and how to design this/these method(s), but I hope you get what I mean. I still see value in providing a very simple to use method like current version of `touch_swipe`, but allow for more advanced interactions as well. I would also be open to ship some default swipe actions in pyatv directly for common patterns.

pyatv/protocols/companion/api.py
postlund: I believe @thiccaxe is correct here. It makes the most sense that the client includes the *actual* timestamp of the event. In reality, it should not make that much of a difference other than possibly some stuttering.

pyatv/protocols/companion/api.py
```diff
@@ -283,6 +291,90 @@ async def hid_command(self, down: bool, command: HidCommand) -> None:
...
+    async def touch_click(self, action: InputAction):
+        """Send a touch click.
+
+        :param action: action mode single tap (0), double tap (1), or hold (2)
+        """
+        if action in [InputAction.SingleTap, InputAction.DoubleTap]:
+            count = 1 if action == InputAction.SingleTap else 2
+            for i in range(count):
+                await self._send_command("_hidC", {"_hBtS": 1, "_hidC": 6})
+                await asyncio.sleep(0.02)
```
postlund, line 365: Is this sleep time based on what you've seen with atvproxy?

pyatv/protocols/companion/api.py
```diff
@@ -283,6 +291,90 @@ async def hid_command(self, down: bool, command: HidCommand) -> None:
...
+                await self._send_command("_hidC", {"_hBtS": 2, "_hidC": 6})
+                await self.hid_event(
+                    int(TOUCHPAD_WIDTH), int(TOUCHPAD_HEIGHT), TouchAction.Click
+                )
+        else:  # Hold
+            await self._send_command("_hidC", {"_hBtS": 1, "_hidC": 6})
+            await asyncio.sleep(1)
```
postlund, line 372: Same question here, since this has another length than above (should they be the same?).

tests/fake_device/companion.py
```diff
@@ -1,12 +1,13 @@
...
 
 import asyncio
+from dataclasses import dataclass
 from enum import IntFlag, auto
 import logging
 import plistlib
 from typing import Any, Dict, List, Mapping, Optional, Set
 
-from pyatv.const import KeyboardFocusState
+from pyatv.const import HidEventMode, KeyboardFocusState
```
postlund, line 10: `TouchAction`

tests/fake_device/companion.py
```diff
@@ -57,6 +58,14 @@
 }
 
 
+@dataclass
+class HidEvent:
+    press_mode: HidEventMode
```
postlund, line 63: `TouchAction`

tests/fake_device/companion.py
```diff
@@ -388,6 +404,50 @@ def handle__hidc(self, message):
...
+    def handle__touchstop(self, message):
+        _LOGGER.debug("Touch stop command received")
+        self.send_response(message, {})
+
+    def handle__hidt(self, message):
+        press_mode: int = message["_c"]["_tPh"]
+        ns = message["_c"]["_ns"]
+        cx = message["_c"]["_cx"]
+        cy = message["_c"]["_cy"]
+        if press_mode == 1:
```
postlund, line 438: You can do something like this:

```python
match press_mode:
    case TouchAction.Press:
        ...
    case TouchAction.Hold:
        ...
```

pyatv/protocols/companion/api.py
```diff
@@ -351,3 +443,25 @@ async def fetch_attention_state(self) -> SystemStatus:
...
         return SystemStatus(content["state"])
+
+    async def _touch_start(self) -> Mapping[str, Any]:
+        """Subscribe to touch gestures."""
+        self._base_timestamp = time.time_ns()
+        response = await self._send_command(
+            "_touchStart",
+            {"_height": TOUCHPAD_HEIGHT, "_tFl": 0, "_width": TOUCHPAD_WIDTH},
+        )
+        self.dispatch("_touchStart", response.get("_c", {}))
```
postlund, line 454: No one is listening to this, so not needed.

pyatv/protocols/companion/api.py
```diff
@@ -351,3 +443,25 @@ async def fetch_attention_state(self) -> SystemStatus:
...
+        return response
+
+    async def _touch_stop(self) -> None:
+        """Unsubscribe touch gestures."""
+        await self._send_command("_touchStop", {"_i": 1})
+
+    async def _siri_start(self) -> Mapping[str, Any]:
```
postlund, line 461: Siri commands does not belong in this PR.

postlund:
Left a few comments, think that is it.

pyatv/protocols/companion/api.py
albaintor: Yes exactly

pyatv/protocols/companion/api.py
albaintor: Here this is hold command, the hold duration can vary but this is the minimal duration so that it will be taken into account

tests/fake_device/companion.py
albaintor: Done

albaintor:
Hi,

I have not added the swipe sequences "templates" as there seem to be several ways to do it. 
Also if we want to cover oblique gestures (from bottom left to top right) it would multiply the templates but I like this idea of bringing some kind of helper methods. 
What could be done instead is to define all the possible possitions : LEFT, RIGHT, TOP, BOTTOM, TOP_LEFT, TOP_RIGHT, BOTTOM_LEFT, BOTTOM_RIGHT and then having a method which takes 2 parameters : start & end coordinates?

thiccaxe:
If singular cardinal directions are wanted through this api you must only move 200 pixels or so in that cardinal direction. You can't guarantee that a longer swipe will move, say, 2-3 ui elements because the Apple TV user can change the input sensitivity in settings. The better options is many smaller swipes-if swipes are absolutely needed. Might as well use the hidC option instead

albaintor:
> If singular cardinal directions are wanted through this api you must only move 200 pixels or so in that cardinal direction. You can't guarantee that a longer swipe will move, say, 2-3 ui elements because the Apple TV user can change the input sensitivity in settings. The better options is many smaller swipes-if swipes are absolutely needed. Might as well use the hidC option instead

The touch range is 1000px x 1000px : so this is true if you are using the apple remote, however from an iphone/ipad you have a bigger touch window so more precise gestures. The final goal is to make touch gestures available from third party devices like home assistant or a remote with a touchscreen

thiccaxe:
> The touch range is 1000px x 1000px : so this is true if you are using the apple remote, however from an iphone/ipad you have a bigger touch window so more precise gestures. 

I agree. However there can be some unexpected behavior as well (writing here for future users of this api). In tvOs a single swipe =/= many small swipes. For example, some ui elements are bounded in a no swipe zone, where you cannot exit the zone as a second motion in a single, continuous swipe, but can only do so as the first motion in a continuous swipe.

thiccaxe:
> The final goal is to make touch gestures available from third party devices like home assistant or a remote with a touchscreen

For the remote situation, I wonder if translating the user's interactions directly to touch packets might be a better solution than another layer of abstraction/api through pyatv.

albaintor:
> > The final goal is to make touch gestures available from third party devices like home assistant or a remote with a touchscreen
> 
> For the remote situation, I wonder if translating the user's interactions directly to touch packets might be a better solution than another layer of abstraction/api through pyatv.

You're right, I think we will have both needs : I am currently working for this project https://www.unfoldedcircle.com/ and both functions will be useful: 

1. Touch swipes horizontal (mostly) and vertical : it will address issues with inconsistent handle of fast forward/rewind among apps on apple TV.
2. Touchpad on the touch screen or with the "airmouse" feature of the remote -> for this one I will use the hid (lowlevel) events directly

postlund:
I believe we have a lot of valid points here, but maybe we should focus on the essentials in this PR and open new PRs with improvements. I will try to review the last few changes and have this merged.

pyatv/core/facade.py
```diff
@@ -601,6 +601,41 @@ def playstatus_error(self, updater, exception: Exception) -> None:
...
+            duration_ms=duration_ms,
+        )
+
+    @shield.guard
+    async def touch_action(self, x: int, y: int, mode: TouchAction) -> None:
+        """Generate a touch event to end x,y coordinates."""
+        return await self.relay("touch_action")(x=x, y=y, mode=mode)
+
+    @shield.guard
+    async def touch_click(self, action: InputAction):
```
postlund, line 631: `click`

postlund (changes requested):
Minor name changes to be consistent with previous change.

postlund:
For later reference... size of the track pad can be saved as settings in storage.

thiccaxe:
On that note, I will test if the trackpad size actually makes a difference, maybe through atvproxy.

albaintor:
All done here !

postlund (approved):
Will merge once checks pass! 👍

postlund:
@albaintor There are a few linting stuff to fix.

albaintor:
> @albaintor There are a few linting stuff to fix.

albaintor:
Hi,
it was sorting issues. Should be fixed now thanks

postlund:
@albaintor I don't think you managed to push any changes?

albaintor:
> @albaintor I don't think you managed to push any changes?

Are you sure ? I have pushed everything on my side
For example : pyatv/interfaces.py:1284 --> method renamed click instead of touch_click (N-1 commit)
And linting fix pushed from https://github.com/albaintor/pyatv/commit/d307a8cdd39aae5388b05c9915c55e9873831dfc
For example : pyatv/core/facade.py
image

postlund:
@albaintor I just noted that there are no pushed changes after my last comment according to the PR. There are also still checks that are failing, you can check that below.

albaintor:
Hi I think I know what is going on : something is wrong on the build action
image

The black command fails with the new "match case" code in tests/fake_device/companion.py
The target python should be added in the black : -t py310 or whatever

postlund:
> Hi I think I know what is going on : something is wrong on the build action image
> 
> The black command fails with the new "match case" code in tests/fake_device/companion.py The target python should be added in the black : -t py310 or whatever
> 
> EDIT : Addin `black --target-version py310 --fast --check .` within chickn.yaml would resolve the problem in order to accept python >3.10 match / case routines. But it also raises new linting issues. Just want to be sure what do you want to do from there See ref [psf/black#2242](https://github.com/psf/black/issues/2242)

Ah, I don't want to drop support for older Python versions just because of this. Just revert to an if-statement instead. You can use `./scripts/chickn.py -t fixup` to deal with linting properly (it's what's used by CI).

albaintor:
> > Hi I think I know what is going on : something is wrong on the build action image
> > The black command fails with the new "match case" code in tests/fake_device/companion.py The target python should be added in the black : -t py310 or whatever
> > EDIT : Addin `black --target-version py310 --fast --check .` within chickn.yaml would resolve the problem in order to accept python >3.10 match / case routines. But it also raises new linting issues. Just want to be sure what do you want to do from there See ref [psf/black#2242](https://github.com/psf/black/issues/2242)
> 
> Ah, I don't want to drop support for older Python versions just because of this. Just revert to an if-statement instead. You can use `./scripts/chickn.py -t fixup` to deal with linting properly (it's what's used by CI).

Hi, this is all done including linting
There are some linting warnings but not related to my modifications

albaintor:
> As long as the linting fails, you must fix it. You can suppress too-many-branches and you probably have to move the existing suppression for logging.TRAFFIC.

It should be okay this time

albaintor:
All tests passed !

postlund:
Yes! 🥳

PR #2412, opened 2024-06-22
Fix Chacha20Cipher with companion

bdraco:
followup to #2338


bdraco:
Thanks. Sorry I missed that on the first go

PR #2466, opened 2024-07-31
opack: implement dates

thiccaxe:
I'd also like to add support for negative integers into opack, if possible. If anyone has a mac, I'd ask if you could test this piece of code https://github.com/fabianfreyer/opack-tools/ to see whether or not opack supports negative integers in the first place. 

My suspicion is that OPACK does not, in fact, have a data type for signed integers, and attempting to encode negative integer will be treated as an unsigned integer internally. (e.g. `OPACKDecode(OPACKEncode(NSNumber(-15))) = NSNumber(4294967281)`, at a high level)

### DIFF
--- a/pyatv/support/opack.py
+++ b/pyatv/support/opack.py
@@ -1,7 +1,6 @@
 """Support for the OPACK serialization format.
 
 Notes:
- * Absolute time (0x06) is not implemented (can unpack as integer, not pack)
  * Pack implementation does not implement UID referencing
  * Likely other cases missing
 """
@@ -14,6 +13,8 @@ from typing import Dict, Tuple, Type
 from uuid import UUID
 
 _SIZED_INT_TYPES: Dict[int, Type] = {}
+
+_OPACK_TIME_OFFSET: float = 978307200.0
 
 
 def _sized_int(value: int, size: int) -> int:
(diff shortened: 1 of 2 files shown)


pyatv/support/opack.py
```diff
@@ -15,6 +14,8 @@
 
 _SIZED_INT_TYPES: Dict[int, Type] = {}
 
+_OPACK_TIME_OFFSET: float = 978307200.0
```
postlund, line 17: Can you add a comment (and possibly rename the constant) making it clear that this is a conversion from offset between unix time and mach absolute time? This is a very magic constant.

tests/support/test_opack.py
```diff
@@ -36,8 +36,7 @@ def test_pack_uuid():
 
 
 def test_pack_absolute_time():
-    with pytest.raises(NotImplementedError):
-        pack(datetime.now())
+    assert pack(datetime.fromtimestamp(978307200.0)) == b"\x06" + (b"\x00" * 8)
```
postlund, line 39: Same with the tests, I don't really like magic constants. I have used mach time in MRP IIRC, so you could add some kind of helper method to convert `datetime` objects mach time (and possibly pack) if you like.

thiccaxe:
I'll add in a conversion function, and use `datetime.timedelta`, as plistlib does

\-\-\-\-\-
After wondering why dates weren't being encoded as dates for about 30 minutes, it hit me like a train that I was looking at bplist decoding and not opack decoding. And not only that, while the field is semantically a timestamp, it is encoded as a float in the bplist. Well, the opack conversion should still be accurate, though there might be a mixup between seconds and milliseconds.

postlund:
I'm not sure where we are with this right now. Can we clean it up and merge it?

Issue #2403, opened 2024-06-21
Verify compatibility with tvOS 18


postlund:
### What to investigate?

Upgrade to current beta and try it out.

### Expected outcome

See if anything needs to be fixed.

knopp:
Video playback (play_url) seems to fail with 500 at `/GET playback-info`. Even when ignoring the error playback doesn't start.

postlund:
When playing video or audio or both?

knopp:
I tried it with a [HLS URL](https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_4x3/bipbop_4x3_variant.m3u8).

```
atvremote -n "Apple TV 4" --debug play_url=https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_4x3/bipbop_4x3_variant.m3u8
```

postlund:
@knopp Could you possibly attach some debug logs I can look at?

postlund:
I can confirm the issue now as well, not sure why it happens. Will have to investigate further.

postlund:
I suspect that the endpoint I'm using (`/playback-info`) is no longer available as that is only for AirPlay 1 and that call should not be made anymore. Not sure yet what to do instead though, will dig into that.

thiccaxe:
If you change the client version in the companion protocol to a i(pad)os18 or higher version, and subscribe to "NowPlayingInfo" events, in some AppleTV+ titles that send over large packets, there is an extraneous null byte inserted before the beginning of the bplist that is encoded as the value of the dictionary key "NowPlayingInfoKey". Removing this null byte allows everything to be decoded properly.

one of more of the following components fail:
- Opack Decoding
- Decryption (no errors)

If this takes place during decryption, which I doubt, the null byte would have to be inserted after the decryption takes place, and somewhere along the way from the crypto internals to the api surface. If it happened before, there would be a mismatch between the data length and the header, which would lead to a decryption error, likely. Moreover, the extra null byte doesn't appear at a power-of-2 location, which could be a internal crypto block size (i.e byte 63/64/65, or byte 127/128/129). This all points to the fact that the encryption is fine

This means that it is likely that the opack decoder is not "up to the standard". It could also be obfuscation measures from apple.

If someone with a mac wants to test the opack, or replicate this "bug?", t o obtain the "bad" opacks, take a look at the following code:
```py
import pyatv
import asyncio, binascii
import logging
from datetime import datetime
import plistlib
[...]

PR #2472, opened 2024-08-07
opack: fix byte encoding length

thiccaxe:
https://github.com/postlund/pyatv/issues/2403#issuecomment-2267340404

### DIFF
--- a/pyatv/support/opack.py
+++ b/pyatv/support/opack.py
@@ -86,17 +86,17 @@ def _pack(data, object_list):
             packed_bytes = (
                 bytes([0x91]) + len(data).to_bytes(1, byteorder="little") + data
             )
-        elif len(data) <= 0xFFFF:
+        elif len(data) <= 0xFFFF:  # 2^16-1
             packed_bytes = (
                 bytes([0x92]) + len(data).to_bytes(2, byteorder="little") + data
             )
-        elif len(data) <= 0xFFFFFF:
+        elif len(data) <= 0xFFFF_FFFF:  # 2^32-1
             packed_bytes = (
-                bytes([0x93]) + len(data).to_bytes(3, byteorder="little") + data
+                bytes([0x93]) + len(data).to_bytes(4, byteorder="little") + data
             )
-        elif len(data) <= 0xFFFFFFFF:
+        elif len(data) <= 0xFFFF_FFFF_FFFF_FFFF:  # 2^64-1
             packed_bytes = (
-                bytes([0x94]) + len(data).to_bytes(4, byteorder="little") + data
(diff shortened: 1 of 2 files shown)


postlund:
I'm not sure this is correct, I will have to verify with macOS implementation first.

postlund:
Nope, you are correct 👍 

```raw
postlund@Pierre ~ % echo 94010000000000000001 | xxd -r -p | ./decode
2024-08-07 19:48:34.307 decode[12510:963399] Decoded: {length = 1, bytes = 0x01}
```

Please update the documentation to reflect this as well.

thiccaxe:
@postlund, I took a look at the docs, and realized that the same might apply to strings and pointers as well.

I think pointers would be hard to test, and I think either way no one will run into a 32 bit or 64 bit pointer... At least strings, though, can you test?

postlund:
> @postlund, I took a look at the docs, and realized that the same might apply to strings and pointers as well.
> 
> I think pointers would be hard to test, and I think either way no one will run into a 32 bit or 64 bit pointer... At least strings, though, can you test?

You are probably right. I'll see if I can try it out. Even if we use a large pointer since, we can still put a small value in the actual payload. So it's not much harder to test than smaller values.

thiccaxe:
Any update?

postlund:
> Any update?

I'll merge it if you update the documentation related to your changes 👍

thiccaxe:
I was actually wondering if you had been able to test strings or pointers for the similar length  behavior.

postlund:
Ah, right. I'll see if I can do that.

postlund:
So, I used my example from the documentation but added an additional element at the end to ensure the pointer is somewhere inside the payload (I added `0=1` to the dict). It looks like this:

```
echo E4416102416244746573744163A20809 | xxd -r -p | ./decode
2024-09-09 20:02:52.341 decode[73452:3341836] Decoded: {
    0 = 1;
    b = test;
    c = test;
    a = 0;
}
```

The idea is now to change the pointer for the `c` key from A2 to something like C4 whilst maintaining the same result as above. Based on the documentation I already have, it should be `C4 02 00 00 00`. So let's try that:

```
echo E4416102416244746573744163C4020000000809 | xxd -r -p | ./decode
2024-09-09 20:05:01.990 decode[73483:3343621] Decoded: {
    0 = 1;
    b = test;
    c = test;
    a = 0;
}
```

So I would say current documentation is correct (although, I see now that C2 Is mentioned twice...). It makes sense for pointers as pointers are just indices and not raw byte position in the payload.

postlund:
I will merge this, we can deal with documentation later!

PR #2552, opened 2024-11-04
Expose delay parameter for Hold action in Companion Protocol

dmrgn1991:
@postlund I realized that I forgot to add any way of exposing the delay parameter for the Hold action in Companion. In practice this means that the delay value is currently hard-coded and as a result the app I'm testing is still uncooperative. This minor change exposes the delay param to the public key methods so that the user can set longer holds. Sorry I didn't think of this sooner, it should have been in my initial PR for sure.

### DIFF
--- a/pyatv/protocols/companion/__init__.py
+++ b/pyatv/protocols/companion/__init__.py
@@ -286,33 +286,33 @@ class CompanionRemoteControl(RemoteControl):
         self.api = api
 
     # pylint: disable=invalid-name
-    async def up(self, action: InputAction = InputAction.SingleTap) -> None:
+    async def up(self, action: InputAction = InputAction.SingleTap, delay: float = 1) -> None:
         """Press key up."""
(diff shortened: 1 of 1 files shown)


postlund:
I have not added this before as I have not seen any direct need for it, but I do see a use case for it so it is ok to add. As all the methods are part of the external interface (as defined in `interface.py`), it's not as easy as just adding new arguments. You must update the interface, update the facade (`facade.py`) and make sure to add the argument to all other protocols as well. Preferably update the other protocols to support the delay too. Delay would only be applicable to `hold` and can have the default value of one second.

I suggest that you read up about relaying and the facade here:

https://pyatv.dev/internals/design#relaying

It should help you get going. It is also worth adding this to the documentation to help others knowing about this feature:

https://pyatv.dev/development/control/

dmrgn1991:
Okay, that makes sense, and I should have realized this would be more involved. I'll look into this and try to provide a complete implementation. Thanks, as always.

PR #2583, opened 2024-12-12
Fix typo, expand acronym at first use

blech:
There was a typo ('vevsion') so I fixed that, then scanned to see if there was anything else worth changing, and came across an acronym that wasn't defined at its first use, so I tweaked that too.


postlund:
Thank you for your contribution, @blech!

PR #2627, opened 2025-02-28
deps: Set cryptography 44.0.1 as base version

postlund:
Earlier versions has a vulnerability so we should not support that.


IsakTheHacker:
Hey @postlund, could you point me to some issue or post where this security vulnerability is documented? I want to know if it could affect me.

Thanks in advance, Isak

PR #2628, opened 2025-02-28
Debug slow tests on windows with python 3.12

postlund:


postlund:
Must have been some issues with runners on GitHub, seems to work now.

Issue #2656, opened 2025-03-18
Connection crash with tvOS 18.4


albaintor:
### Describe the bug

I have 2 AppleTV boxes : one on tvOS 18.3 and another with 18.4 beta.
The one with 18.4 beta raises a connection lost exception right after the connection (the other one works fine).
I paired the device with 2 protocols : companion and airplay. The disconnection is raised when using the companion protocol, (airplay is fine)

See the sample code below to reproduce it with the following steps
1. Pair the Apple TV device with companion protocol using tvOS 18.4
2. Connect to it with pyatv
3. Subscribe to connection lost/closed events by implementing the `DeviceListener` interface
4. The client will receive a lost connection event a few milliseconds later

```python
class AppleTv(DeviceListener):

    async def start_client:
        self._atv = AppleTv(device=config, loop=_LOOP)
        await self._atv.connect()
        self._atv.listener = self
        

    def connection_lost(self, _exception) -> None:
        _LOG.exception("[%s] Lost connection %s", self.log_id, _exception)
        self._handle_disconnect()

    def connection_closed(self) -> None:
        _LOG.debug("[%s] Connection closed!", self.log_id)
        self._handle_disconnect()


### Error log

Extraction of logs (full logs attached), see the `pyatv.protocols.companion.connection:Connection lost to remote device: None`
```log
DEBUG:__main__:[Salon Apple TV] Updating app list
DEBUG:pyatv.protocols.companion.protocol:Exchange OPACK: {'_i': 'FetchLaunchableApplicationsEvent', '_t': 2, '_c': {}, '_x': 20683}
DEBUG:pyatv.protocols.companion.protocol:Send OPACK: {'_i': 'FetchLaunchableApplicationsEvent', '_t': 2, '_c': {}, '_x': 20683}
DEBUG:pyatv.protocols.companion.connection:>> Send data (Data=e4425f696046657463684c61756e636861626c654170706c69636174696f6e734576656e74425f740a425f63e0425f7831cb50, FrameType=08)
DEBUG:pyatv.protocols.companion.connection:>> Send (Encrypted=9321a4c7e7672125d7db2a4e7e1e1085ad6c605d165f5368a476b89b75d326b14f4a7d4a3b2221343d9cbee4ef0d264de86dcda9f2b08497c63b6509afaf1fd60bc865, Header=08000043)
[...]

PR #2657, opened 2025-03-19
Fix (temporary?) for tvOS 18.4 for companion protocol disconnection

albaintor:
Fix for https://github.com/postlund/pyatv/issues/2656
If confirmed that Apple has blacklisted the hardcoded previous ID `cafecafecafe` it should be a definitive fix.


postlund:
Sorry for the lack of response, I will look into this. I believe a random value is probably better than we have, but I want to investigate a bit further first. I believe _i is some kind of device ID and if it's important, then maybe a random one should be generated and saved among settings so it can be re-used for future connections.

albaintor:
Yes it is certainly related to the device as it seems to be constant over time.
But the more I think about it, more I believe that Apple blacklisted your static value, as any other values work. I don't understand why, what is the harm with this library ? this looks kinda mean from Apple policy if confirmed

postlund:
> No problem, we all have a life ! Yes it is certainly related to the device as it seems to be constant over time. I tried to look and convert some device ids in hexa but I didn't find anything close. But the more I think about it, more I believe that Apple blacklisted your static value, as any other values work. I don't understand why, what is the harm with this library ? this looks kinda mean from Apple policy if confirmed.

I have some leads, we'll see what I can dig up tomorrow. Hopefully something useful. Would be very strange if they black listed that. If you pick a very close ID, just changing like a single octet, that still works?

> 
> About storing the id in order to reuse it later, it means that it will be returned with pairing key at the end, and would be an optional parameter for connection (and if not provided generate a random one), something like that ?

I would put a new field in InfoSettings and store in as part of the storage model. Since it's a device identifier of some sort, other protocols could potentially need it so it makes sense to store it on "device level". Question is what to call it since we don't know what it is yet...

https://github.com/postlund/pyatv/blob/a6cb277e895c42f7d65d09abe9bf08aa5682e928/pyatv/settings.py#L76

albaintor:
Ok
Yes if you submit cafecafecaf0 or anything close it works. 
Unless.... maybe having several clients connecting with different keys but same id cafe...

postlund:
@albaintor Yes, I do believe we should generate a unique identifier and re-use that. It seems to be device specific from what I can see. I'll draft something.

postlund:
Will be fixed by #2680

PR #2673, opened 2025-04-08
Added ability to set volume per (connected) device

albaintor:
Actually the volume set is only done on the AppleTV side. But when you set your AppleTV to cast audio to multiple devices (homepod or any Airplay speaker), the set volume won't affect the other devices.

I added a new endpoint to specify the endpoint id while setting volume, in order to be able to reproduce the global volume slider from the notification center (it lets change all volumes at the same time)


albaintor:
EDIT : I also added a new event to be notified when the volume of output devices changes

postlund:
Thank you, I will review this later tonight. But you have to revert the changes related to protobuf.

albaintor:
Hi,

it's done and tests passed

pyatv/protocols/companion/api.py
```diff
@@ -196,7 +197,7 @@ async def system_info(self):
                 "_bf": 0,
                 "_cf": 512,
                 "_clFl": 128,
-                "_i": "cafecafecafe",  # TODO: Figure out what to put here
+                "_i": os.urandom(6).hex(),
```
postlund, line 200: This is out of scope of this PR.

pyatv/interface.py
```diff
@@ -1174,6 +1193,14 @@ async def set_volume(self, level: float) -> None:
         """
         raise exceptions.NotSupportedError()
 
+    @feature(66, "SetDeviceVolume", "Set device volume level.")
```
postlund, line 1196: I suggest that we don't add a new method for rather but rather re-use `set_volume` and just add an optional output device as argument instead:

```python
    async def set_volume(self, level: float, output_device: Optional[OutputDevice] = None) -> None:
```

pyatv/interface.py
```diff
@@ -1584,3 +1611,22 @@ def keyboard(self) -> Keyboard:
     @abstractmethod
     def touch(self) -> TouchGestures:
         """Return touch gestures interface."""
+
+
+class OutputDeviceState:
```
postlund, line 1616: `interface.py` is supposed to only contain public API stuff and this is purely internal. I believe `core/__init__.py` is a more suitable place. Feel free to make it a data class instead. It's not widely used within pyatv for historical reasons, but my intention is to move towards that.

pyatv/interface.py
```diff
@@ -1140,6 +1152,13 @@ def volume_update(self, old_level: float, new_level: float) -> None:
         """Device volume was updated."""
         raise NotImplementedError()
 
+    @abstractmethod
+    def volume_device_update(
+        self, output_device_id: str, old_level: float, new_level: float
```
postlund, line 1157: Change to `OutputDevice` instead of a device Id string.

pyatv/interface.py
```diff
@@ -1121,6 +1124,15 @@ def identifier(self) -> str:
         """Return a unique id for the output device."""
         return self._identifier
 
+    @property
+    def volume(self) -> float:
+        """Return the current volume of the output device."""
+        return self._volume
+
+    @volume.setter
```
postlund, line 1132: This class is supposed to be immutable (that is generally for data classes in the public API), so a setter here won't work here.

postlund (changes requested):
Looks good overall, but a few comments to make conform with how the API generally work in pyatv. I would also like to see some kind of test (unit test in facade for instance) since this is new functionality.

pyatv/interface.py
albaintor: Ok done (dataclass)

pyatv/interface.py
albaintor: Done

pyatv/protocols/companion/api.py
albaintor: Removed

pyatv/interface.py
albaintor: I have to find something out : currently only the output devices can be updated added/removed
Their attributes need to be updated (volume actually)

albaintor: I changed this class to a dataclass but had to keep it in interfaces.py (otherwise circular import). Also this does not change the issue you reported. Using dataclass allow modifications too.

I don't see how it coud be done differently : this class `OutputDevice` needs to be updated (volume level).

albaintor:
Hi again, I have added a test for device volume update and tested ok.

There is one thing that I changed partially : you requested OutputDevice class in the interfaces.py to be immutable.
I didn't manage to do this, I just migrated the class to a dataclass but this does not address the request.

To be honest I don't know how to do this in an easy way : the volume property needs to be updated.

albaintor:
Hi @postlund did you have the time to look after the recent changes ? thank you

albaintor:
Hi @postlund any news here ? thank you

postlund:
@albaintor Can you take a look at this because I believe you have changed line endings or something. All files more or less as marked as changed. I would also like you to remove the "tvOS 18.4" commit as that has been merged already.

albaintor:
Hi, I don't know what happened with CRLF... this is fixed. I also removed the tvOS 18.4 patch.
There is still the thing I didn't manage to handle, mentioned in this post https://github.com/postlund/pyatv/pull/2673#issuecomment-2796287243

About using immutable data whereas I need to send updated data (updated volume)

albaintor:
Hi @postlund , did you have the time to look after my PR ? thank you

PR #2680, opened 2025-04-11
companion: Use static id for _i in _systemInfo

postlund:
This change generates a random "rp identifier" that is included in _systemInfo sent by the Companion protocol. It's not really known what this identifier is, but it comes from rapportd/Rapport.framework and seems to be unique per device. So we generate a unique one per device and store it with settings.

Adjustments have been made so that if this identifier is missing in settings, a new identifier will be generated and saved automatically.

Relates to #2656

### DIFF
--- a/docs/api/pyatv.const.html
+++ b/docs/api/pyatv.const.html
@@ -950,19 +950,19 @@ is Apple TV Software (pre-tvOS).</p></section>
 <dl>
 <dt id="pyatv.const.TouchAction.Click"><code class="name">var <span class="ident">Click</span> = 5</code></dt>
 <dd>
-<section class="desc"><p>The type of the None singleton.</p></section>
+<section class="desc"></section>
 </dd>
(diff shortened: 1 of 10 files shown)


postlund:
@albaintor I added a new setting per device, `rp_id`, that contains a random identifier. It it is stored persistently. Can you verify if this works?

albaintor:
Hi @postlund ,

I just tested it and it works well so far
Thanks

Ghawken:
Hi,

Running this pull currently as well on one 18.4 AppleTV.

Seeing repeated connection lost. 

Works on restart - but then connection lost as logging here recurs.

(
[Log.zip](https://github.com/user-attachments/files/19722605/Log.zip)

Not completely sure its not my end - but thought would post logging info here for you)

Thanks again!

jeffjumper:
Could this be merged into a release? At the moment, every time there's a new Home Assistant release, I need to ssh into the HA docker instance to hand-edit the api.py file and change cafecafecafe to a random identifier or my log quickly fills up with connection errors.

PR #2735, opened 2025-07-13
Use asyncio.timeout for Python 3.11+

cdce8p:
`async-timeout` is deprecated for Python 3.11+ as it's now included in the stdlib.
https://github.com/aio-libs/async-timeout


postlund:
Thank you @cdce8p! Can you fix the conflict? May I suggest that you move the conditional import code to `pyatv/support/__init__.py` and just do `from pyatv.support import async_timeout` instead? Maybe also add a comment/todo there that we can remove that part once the lowest supported python version is 3.11?

PR #2736, opened 2025-07-14
Drop support for pydantic v1

cdce8p:
Pydantic will only support Python 3.14 for `v2` models. Home Assistant has since also moved to `v2`, so it's likely safe to drop support for `v1`. https://github.com/pydantic/pydantic/issues/11613#issuecomment-3061291322


edenhaus (approved):
I have tested this PR with Home Assistant on Python 3.13.9 and 3.14.2 and it works without any issues with the Apple TV 4K

PR #2791, opened 2025-12-05
fix(net): fallback multicast join without adapters

Skyline-23:
This pull request improves network interface detection and multicast socket setup for sandboxed platforms where
  ifaddr may return no adapters. When adapters are present, behavior is unchanged. The fallback uses a dummy UDP
  connect (no traffic sent) solely to learn the default local IPv4.

  ### Network interface detection improvements

  - get_private_addresses now probes the default IPv4 via a dummy UDP connect when no adapters are reported, keeping
    loopback filtering rules.

  ### Multicast socket setup enhancements

  - mcast_socket falls back to the probed default IPv4 for IP_MULTICAST_IF/IP_ADD_MEMBERSHIP, with a final 0.0.0.0
    membership attempt if probing fails.
  - Binding remains "", avoiding port conflicts while only multicast membership targets the detected address.


pyatv/support/net.py
```diff
@@ -47,6 +47,26 @@ def mcast_socket(address: Optional[str], port: int = 0) -> socket.socket:
         with suppress(OSError):
             membership = socket.inet_aton("224.0.0.251") + socket.inet_aton(address)
             sock.setsockopt(socket.IPPROTO_IP, socket.IP_ADD_MEMBERSHIP, membership)
+    else:
+        # Fallback for sandboxed platforms where ifaddr finds no adapters:
+        # detect default IPv4 via a dummy UDP connect and use it for multicast join.
+        try:
+            with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as probe:
+                probe.connect(("8.8.8.8", 80))
```
postlund: Can we accomplish this without hardcoding a public IP address? Using pyatv should not require access to Internet.

Skyline-23:
Applied fix to avoid public IP for route probe; pushed commit ae6e10e2 to Skyline-23:master.

PR #2834, opened 2026-03-17
Reverse-engineered MRP protobuf definitions from Apple shared libraries

JarekToro:
### Context

While reverse-engineering Apple's shared frameworks (MediaRemote.framework
and related libraries on macOS/iOS) via Ghidra, I reconstructed a much
larger set of MRP protobuf schemas than what currently exists in pyatv.
The RE work was done as part of building [bunatv](https://github.com/JarekToro/bunatv).

pyatv's thorough work and documentation were invaluable to me, so I wanted to share everything I found back with the
project. This PR isn't meant to be merged as-is, these definitions are
yours to use however makes sense, whether that's cherry-picking specific
messages, using them as reference, or adapting the structure.

While I can attest that within the latest apple shared libs  _every_ proto message here has specific read and write functions for converting to and from protobuf and native classes,  **I don't know** how various messages are meant to be used, and if they are all meant for apple Tvs, and if they are supported by all devices. 

If this proves useful, I've also started applying the same RE approach to the Companion protocol,
though it's slower going, the relevant code is spread across significantly
more shared libraries than MRP. Happy to share those findings too
as they come together.

> AI tools were used during the decompilation process. I've cleaned up the
> output, but there may be a few residual comments from that process that
> are specific to the decompilation context rather than generally useful.


kristofferR:
Damn, this is cool!

snoslicer8:
Noting here that I have a gen 3 Apple TV 4K on tvOS 26.5 where FetchAttentionState gets no response, while a gen 2 on the same tvOS works fine. If it helps your RE efforts.

PR #2846, opened 2026-04-25
Fix compatibility with latest AirPlay

jlacivita:
## Summary
- Fixed `play_url` to work with the latest AirPlay protocol changes

## Test plan
- [ ] Manually tested `play_url` against AppleTV and Samsung Frame 2025

### DIFF
--- a/base_versions.txt
+++ b/base_versions.txt
@@ -1,4 +1,5 @@
 aiohttp==3.8.3,<5
+biplist==1.0.3
 async-timeout==4.0.2;python_version<'3.11'
 cryptography==44.0.1
 chacha20poly1305-reuseable==0.13.2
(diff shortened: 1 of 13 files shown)


robkochman:
I reworked this approach against current `master` and published the result at
[`robkochman:codex/play-url-tvos26`](https://github.com/robkochman/pyatv/tree/codex/play-url-tvos26),
commit [`8144c77c`](https://github.com/robkochman/pyatv/commit/8144c77c).

The revised implementation keeps the modern `/command` queue flow while:

- using stream-specific identifiers and headers;
- driving playback completion from event-channel state transitions;
- waiting for a real startup transition instead of treating the first state as success;
- avoiding the unrelated dependency churn and raw protocol logging; and
- adding regression coverage for the queue commands and playback state handling.

Validation completed:

- full suite: **1255 passed, 6 skipped**;
- Black, isort, Flake8, mypy, and Pylint all pass;
- live tvOS 26.5 tests: remote HLS, local HLS, and local progressive MP4 all reached
  `playing`; local media then reached `stopped`/`ended` normally; and
- an H.264/AAC visual announcement with embedded audio completed successfully through
  an application wrapper.

I do not have permission to update this PR's source branch directly, so I am linking
the tested branch for the author or maintainers to compare or cherry-pick.

PR #2847, opened 2026-04-28
Fix regression with tvOS 26.5

albaintor:
Fix for https://github.com/postlund/pyatv/issues/2845
To be confirmed if it still works with previous tvOS releases

### DIFF
--- a/pyatv/protocols/companion/api.py
+++ b/pyatv/protocols/companion/api.py
@@ -153,6 +153,7 @@ class CompanionAPI(
         await self.system_info()
         await self._touch_start()
         await self._session_start()
+        await self._tv_rc_session_start()
         await self._text_input_start()
 
         await self.subscribe_event("_iMC")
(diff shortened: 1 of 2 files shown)


albaintor:
I also updated fakedevice for testings

kroehre:
Verified this fixes tvOS 26.4 too 👍

albaintor:
Hi @postlund  can you check this please this affect more and more users

postlund:
I'll try to get this sorted out during the day!

pszypowicz:
Confirming this fixes the regression on Apple TV 4K (2nd gen, AppleTV11,1) running tvOS 26.5 (23L471). Power on/off, app list, and source selection all work again with this branch. Worth noting since issue #2845 reasoned that AppleTV11,1 was unaffected - it isn't, at least on 23L471. Thanks for the fix!

Locutus8:
media_player.turn_on works for me.  But, media_player.turn_off does not.  As a work around, remote.send_command suspend still works for powering off.  Also, you are now able to get a playing state from the media player when the Apple TV is actually off.  I do not believe this is the previous behavior.  The new behavior is the correct behavior as it is possible for the Apple TV to play content when the light is not on and HA should be able to detect and trigger on this.  I assume this part is a change in tvOS 26.5.

kroehre:
@postlund anything I can do to help move this forward? This fix is a key piece to having great apple tv keyboard support in home assistant. Thanks!

michaelmedellin:
@albaintor I opened #2855 proactively. I discovered the same TVRCSesssionStart gap you identified, but also found that I was unable to get fresh connections to receive power-state events at all — pyatv currently sends a null `_i` in `_systemInfo`, and the Apple TV only pushes SystemStatus/TVSystemStatus events to a client whose `_i` is not `null`.  

Have you seen the same issue with no power state events being received by the pyatv client after the session starts?

ErykJas:
It breaks connection to tvOS version `17.6`. During connect it's throwing `pyatv.exceptions.ProtocolError: Command failed: No request handler` exception. This MR: https://github.com/postlund/pyatv/pull/2855 works fine due to wrapping start of RC session into try/catch block for not supported tvOS versions.

postlund:
Looks good, I haven't tried it but I trust you guys! Thanks!

postlund:
Hoooold your horses! #2855 does the same thing but seems more complete to me, I'll roll with that one instead.

Issue #2853, opened 2026-05-18
Keyboard Focus state changed not working


samuellugnani:
### Describe the bug

Focus state changed event is not working on pyatv version 0.17.0, but it's working correctly in version 0.16.1.

### Error log

```log
<place log here>
```

### How to reproduce the bug?

To reproduce the bug, just follow the instructions provided at the link: [https://pyatv.dev/development/listeners/#keyboard-updates](https://pyatv.dev/development/listeners/#keyboard-updates) using pyatv version 0.17.0.

### What is expected behavior?

Apple TV does not send a notification when you move the cursor in and out from a textbox, for example when you search a content in the Search app.

### Operating System

Windows

### Python

3.14

### pyatv

0.17.0

### Device

Apple TV 4K, tvOS 26.5

### Additional context

The event is fired correctly in version 0.16.1

PR #2854, opened 2026-05-18
FIx for Keyboard Focus state changed not working

samuellugnani:
Fix for #2853 
Need to be tested on previous tvOS.

On pyatv version 0.16.1 with pydantic v1 `@field_validator("rp_id", always=True)` made the validator run even when rp_id was not set, so pyatv generated a random rp_id.
On version 0.17.0 with pydantic v2, the always=True has been correctly removed, however that validator is not called on the default None, so when sending _systemInfo message, the "_i" could be None and, after that, no keyboard focus state changes are fired.

With this fix, the default_factory generates the rp_id if it's missing and the validator works only if the rp_id is None.


postlund:
Thank you! 👍

PR #2598, opened 2025-01-09
favor companion protocol for power state and listen to TVSystemStatus events

tinkerborg:
This change uses the OVERRIDE_PRIORITIES in FacadePower throughout the class, and also subscribes to TVSystemStatus events as mentioned in #2325.

With this change I am now able to see power events in all the included scripts, as well as in home-assistant and my own test script. This should resolve the issue noted in #1667 and possibly several others. 

fixes #1667

PR #2855, opened 2026-05-19
Fix Companion Protocol Power State Handling

michaelmedellin:
Fixes how the Companion protocol fetches and subscribes to power state by (1) starting a `TVRCSessionStart` session during connect and (2) sending a non-null `_i` in `_systemInfo`.

## Problem

After #2598 made Companion the preferred power state source (v0.16.1), power state stopped working reliably.

On a new connection — e.g. a full Home Assistant restart — the initial `FetchAttentionState` succeeds, so power state is briefly correct. Later `FetchAttentionState` calls stop being answered, and `CompanionPower.initialize` logs:

```
Could not fetch SystemStatus, power_state will not work (...)
```

After that, power state never recovers. `TVSystemStatus` events are also never pushed.

## Cause

Found by proxying an iOS Remote client against the same Apple TV and replaying its message sequence with single messages dropped:

- Without a TV Remote Client session, the Apple TV stops answering `FetchAttentionState` **and** does not honor the `SystemStatus`/`TVSystemStatus` event subscriptions. We confirmed this by moving past the failed fetch to see whether power state events would still be published — they weren't. That isolated the missing `TVRCSessionStart`, which pyatv never sent.
- Even with the session, the Apple TV only pushes `SystemStatus`/`TVSystemStatus` events to a client whose `_systemInfo` carried a non-null `_i`. pyatv sent `_i: info.rp_id`, which is `None` when no Rapport ID is known.

## Changes

- Send `TVRCSessionStart` after `_sessionStart` during connect. Wrapped in try/except so older devices that reject it are unaffected.
- Fall back to a stable identifier (`device_id` with colons stripped) when `rp_id` is `None`, so `_i` is never null.
- Add functional tests for both, plus fake device support.
- Document `TVRCSessionStart` and the `_i` requirement in `protocols.md`.

## Testing

- Reproduced the proxy/replay bisection that isolated each cause.
- Verified end to end on Home Assistant: power state now tracks `off`/`idle`/`playing`.
- All Companion tests pass; the two new tests fail without the fix.

Related: #2845, #2847

### DIFF
--- a/docs/documentation/protocols.md
+++ b/docs/documentation/protocols.md
@@ -950,6 +950,24 @@ Once a command has been called, it will be cached making it possible to call it 
 sending `_sessionStart` again. This is probably why `_launchApp` keeps working after
 requesting the list from Shortcuts (as it will set up a new session).
 
+A client should also register a *TV Remote Client* session by sending
+`TVRCSessionStart` after `_sessionStart`. Some Apple TVs will not answer commands
+such as `FetchAttentionState` until this session has been registered:
+
+```javascript
+{
+    '_i': 'TVRCSessionStart',
+    '_x': 123,
+    '_t': 2,
+    '_c': {
+        'ProtocolVersionKey': '1.2'
+    }
+}
+```
+
+The response echoes the negotiated `ProtocolVersionKey`. Older devices that do not
+implement this message return an error, so it is safe to always send it.
+
 #### Events
 
 It is possible to subscribe to events using `_interest`:
@@ -966,8 +984,10 @@ It is possible to subscribe to events using `_interest`:
 }
 ```
 
-No explicit response is sent to the request, other than an event update. So far `_iMC`
-(Media Control) is the only known event type. An event update might look like this:
+No explicit response is sent to the request, other than an event update. Known event
+types include `_iMC` (Media Control) and `SystemStatus`/`TVSystemStatus` (power state).
+The Apple TV only pushes `SystemStatus`/`TVSystemStatus` events to a client that sent a
+non-null `_i` in its `_systemInfo`. An event update might look like this:
 
 ```javascript
 {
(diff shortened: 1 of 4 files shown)


QuanTrieuPCYT:
This PR fixed my months-long issue in Home Assistant! Looking forward to getting this merged

forsethc:
Thanks so much, I was chasing a bug with power handling in my project for weeks and couldn't nail it down.  This fixed it.

Please oh please merge it 🙏

MSL-DA:
@postlund Some feedback from the maintainer would be appreciated.

postlund:
Nice! 👍

rafieshaki:
i works thanks 🙏🏼🙏🏼

holnburger:
Thank you for fixing this! Do you already know when a new release of pyatv that includes this fix is happening @postlund?

chemelli74:
> Thank you for fixing this! Do you already know when a new release of pyatv that includes this fix is happening @postlund?

Build failed

postlund:
I have a few more PRs to review and merge before I will make a release (as well as the mentioned build error to take care of).

bbustin:
> I have a few more PRs to review and merge before I will make a release (as well as the mentioned build error to take care of).

Thank you.

When this is released, my automation to limit my kids’ TV time will work again. it is nice to not be the villain. The AppleTV just powers off. If they turn it back on, the second they start watching something it shuts off again. :)

I appreciate everyone’s hard work.

MSL-DA:
It's been a long time coming!

brent20:
Hello! I just wanted to check on the status of a new release with this fix. It's affecting many people using Home Assistant, and I understand these things can take some time, so no worries - I just wanted to follow up. 

https://github.com/home-assistant/core/issues/171666

Thanks!

brent20:
I just want to say thank you!! My partner is no longer upset that Youtube TV doesn't lanuch automatically via Home Asssistant. Thank you @postlund!!

PR #2857, opened 2026-05-25
Preserve underlying error when InternetSource fails to decode

balloob:
## Summary

When streaming over HTTP via RAOP, the underlying error that caused a stream to fail was being swallowed. Users only saw `miniaudio.DecodeError: ('failed to init decoder', -1)` with no hint of the real cause (HTTP 401, connection error, timeout, etc.).

Two issues in `pyatv/protocols/raop/audio_source.py`:

1. **`PatchedIceCastClient._stream_wrapper`** captured download exceptions as `str(ex)` and logged them at DEBUG. The actual exception type and traceback were lost.
2. **`InternetSource.open`** checked `source.error_message` immediately after `miniaudio.stream_any` raised `DecodeError`, racing the download thread. If miniaudio failed before the download thread had recorded its error, the real cause was discarded and only the `DecodeError` propagated.

## Changes

- Store the actual exception (`source.error: Optional[BaseException]`) instead of a string, and log it at WARNING.
- On `DecodeError`, explicitly close the source first (joining the download thread) so any HTTP/network error is captured before we inspect it.
- Raise `ProtocolError(...) from source.error` so the original cause is preserved in the traceback chain.

If the URL truly returns valid audio that miniaudio can't decode, the `DecodeError` is still re-raised unchanged.

## Motivation

This helps diagnose issues like home-assistant/core#169164 where the actual cause (e.g. an HTTP error from the media source) was masked by a generic `failed to init decoder` message.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


postlund:
@balloob flake8 issue is flaking out

pyatv/protocols/raop/audio_source.py
```diff
@@ -607,8 +607,14 @@ async def open(
                 ),
             )
         except miniaudio.DecodeError as ex:
```
NorbertHD: @balloob 
Could you please change this to:

except miniaudio.DecodeError:

So that flake8 doesn't report a lint error.

Quentame: To fix flake8:
```suggestion
        except miniaudio.DecodeError:
```

pyatv/protocols/raop/audio_source.py
```diff
@@ -606,9 +606,15 @@ async def open(
...
-                raise ProtocolError(source.error_message) from ex
+        except miniaudio.DecodeError:
+            # Make sure the download thread has finished so any HTTP/network error
+            # it raised is captured in source.error before we inspect it. Otherwise
+            # we'd race the thread and re-raise the (less informative) DecodeError.
+            await loop.run_in_executor(None, source.close)
+            if source.error is not None:
+                raise ProtocolError(
+                    f"Failed to stream from {url}: {source.error}"
+                ) from source.error
```
Quentame, line 617: Wait, getting confused ... `source.error` is already printed on line 513, right ?

So here we should `raise from DecodeError`, as the comment on line 612 is telling, right ? 😅 

(That also fix flake8)
Sorry 😁 

```suggestion
        except miniaudio.DecodeError as ex:
            # Make sure the download thread has finished so any HTTP/network error
            # it raised is captured in source.error before we inspect it. Otherwise
            # we'd race the thread and re-raise the (less informative) DecodeError.
            await loop.run_in_executor(None, source.close)
            if source.error is not None:
                raise ProtocolError(
                    f"Failed to stream from {url}: {source.error}"
                ) from ex
```

balloob:
@postlund could we get a release with this fix?

pyatv/protocols/raop/audio_source.py
balloob: No, this was reverting the change. We want to raise the other error, as that is the actual issue.

PR #2887, opened 2026-06-30
facade: wire power listener to all protocol implementations

volodymyrpavlenko:
## Summary

- Previously only the main (highest priority) power implementation had its listener set to `FacadePower`. When Companion is the main protocol but fails to track power state (e.g. `FetchAttentionState` times out on newer tvOS), MRP power events were silently dropped because `MrpPower` had no listener wired up.
- Set `FacadePower` as the listener on every registered power implementation so events from any protocol reach the consumer.
- Added test verifying power events from both MRP and Companion reach the facade listener.

## Context

On newer tvOS versions, `FetchAttentionState` may time out or be rejected during companion protocol initialization. The 0.18.0 fix ([`5ed97a3`](https://github.com/postlund/pyatv/commit/5ed97a39)) correctly moved event subscriptions outside the initial fetch try-block, but even when companion event subscriptions also fail, MRP still detects power state changes. However, those MRP events never reached the facade listener because `FacadeAppleTV.connect()` only called:

```python
power = self._interfaces[interface.Power].main_instance
power.listener = self._interfaces[interface.Power]
```

With `FacadePower.OVERRIDE_PRIORITIES = [Companion, MRP, ...]`, `main_instance` always returns `CompanionPower`, leaving `MrpPower.listener` unset. The `StateProducer._ListenerProxy` silently no-ops when no listener is set, so MRP power events are swallowed.

## Test plan

- [x] Added `test_power_events_forwarded_from_all_protocols` — registers both MRP and Companion power, simulates events from each, asserts both reach the facade listener
- [x] All 48 existing `tests/core/test_facade.py` tests pass

### DIFF
--- a/pyatv/core/facade.py
+++ b/pyatv/core/facade.py
@@ -773,14 +773,12 @@ class FacadeAppleTV(interface.AppleTV):
 
         self._device_info = interface.DeviceInfo(devinfo)
 
-        # Forward power events in case an interface exists for it
-        try:
-            power = cast(
-                interface.Power, self._interfaces[interface.Power].main_instance
-            )
-            power.listener = self._interfaces[interface.Power]
-        except exceptions.NotSupportedError:
-            _LOGGER.debug("Power management not supported by any protocols")
+        # Forward power events from all protocol implementations to the facade
+        # so that events are not lost when a non-main protocol detects a change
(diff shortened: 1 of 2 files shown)


volodymyrpavlenko:
This was actually not needed since MRP is noisy and fixing companion in HA helped

Issue #1419, opened 2021-10-23
Airport Express 2 paired fine, but doesn't update the state


Post404:
### Describe the bug

Pairing and everything works as expected with my two airplay express 2.
But the state is always Idle even though i'm playing music from my iPad (iOS 15.0.2) from Spotify.
I just re-added them both to see if that worked but nothing changed.

### Error log

```log
there is no error
```

### How to reproduce the bug?

Add an airplay express 2 device and play music from it.

### What is expected behavior?

That the state would change from idle to playing when playing something.

### Operating System

Home Assistant OS 6.5

### Python

3.9

### pyatv

0.9.5

### Device

Apple Airport Express 2

### Additional context

-

Post404:
Thanks for explaining. I really hope it will be added, because the only way I have found so far is through Homebridge.
Which I ideally want to get rid of.
I was also tinkering around with this I found:

postlund:
I will most certainly try to fix that, it seems like an interesting challenge. But requires some internal changes as mentioned, so it's not a quick fix for it.

Post404:
Alright thanks!

postlund:
@ericvoshall Unfortunately no, I have been quite absent from pyatv development lately. Mainly due to lack of time (family). Hopefully I can take up development soon again, there's quite a lot that needs to be done.

JensTimmerman:
Maybe the airplay protocol doesn't indeed share any of this information, but the airport express certainly does.

```
python3 -c "import urllib.request,plistlib;
   ip='192.168.1.70';
   data=urllib.request.urlopen(f'http://{ip}:7000/info',timeout=2).read();
   p=plistlib.loads(data);
   flags=int(p.get('statusFlags',0));
   print('ON' if (flags & 0x800) else 'OFF')"
```

Works for me to detect if an airport extreme is playing something or not.

Is this something you would be ok to incorporate in pyatv, or should I go directly to the HA's apple tv integration to implement it there?

JensTimmerman:
```
>>> from pyatv.support.rtsp import RtspSession
>>> from pyatv.support.http import http_connect
>>> conn = await http_connect("192.168.1.70", 7000)
... rtsps = RtspSession(conn)
... info = await rtsps.info()
... print('ON' if (info['statusFlags'] & 0x800) else 'OFF')
... 
ON
>>> info = await rtsps.info()
... print('ON' if (info['statusFlags'] & 0x800) else 'OFF')
... 
OFF
```

[...]

PR #2898, opened 2026-07-29
Added zereoconf flag status updates

JensTimmerman:
Started working on a solution for #1419


JensTimmerman:
Need to add some tests, the handler in the servicebrowser doesn't seem to get triggered, need to debug

